PR #11, opened 2023-07-31
Add variables for gross CO2 emissions

danielhuppmann:
This PR adds the variables for gross CO2 emissions (only counting the positive emissions, not any CO2 removal in a sector) developed in the NAVIGATE project.

This variable is the complement to the variables "Carbon Capture" with fossil fuels (to be added in a subsequent PR).

For discussion: do we add the variable tree related to carbon capture and gross-emissions from NAVIGATE directly, or are there any changes that should be made as a clean-up or for consistency?

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -1,6 +1,11 @@
 - Emissions|{Tier-1 Species}:
     description: Emissions of {Tier-1 Species}
     unit: "{Tier-1 Species}"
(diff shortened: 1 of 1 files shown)


strefler:
I would rather say these variables are the complement to Carbon Removal, right? 
Regarding those and the carbon capture variables, I remember a discussion with Ken Oshiro, and they also developed these in one of their projects, and I liked their apporach. So maybe we could discuss with him first before adding the Navigate variables.

strefler (approved):

PR #13, opened 2023-08-01
Include suggested changes from Henry Liu / Chritoph Bertram

byersiiasa:
Issues:

- duplicates mostly between INDIA+, CHINA+ and other Asia
- the 3 baltic countries part of Europe or REF? 

Cambodia
Laos
Mongolia
Vietnam
Bangladesh
Bhutan
Maldives
Nepal
Pakistan
Sri Lanka

PR #14, opened 2023-08-02
Harmonize country mappings

danielhuppmann:
This PR re-implements the region-clean-up (#13), ensuring consistency across R5, R9 and R10 country mappings.

The following design choices were made:
- The Baltic States are part of Europe/EU
- The R9 regions *India* and *China* should include only those countries
- The R10 Region *India+* and *China+* should include neighboring countries that are similar, i.e., on the Indian sub-continent or fit the description "Centrally-planned Asia" (and are close to China)

Of course, these mappings are only a guidance, because actual model implementations will only be an imperfect approximation to these mappings.

### DIFF
--- a/definitions/region/common.yaml
+++ b/definitions/region/common.yaml
@@ -146,10 +146,9 @@
         Democratic Republic of the Congo, Djibouti, Egypt, Equatorial Guinea, Eritrea,
         Ethiopia, Gabon, Gambia, Ghana, Guinea, Guinea-Bissau, Kenya, Lesotho, Liberia,
         Libya, Madagascar, Malawi, Mali, Mauritania, Mauritius, Mayotte, Morocco,
-        Mozambique, Namibia, Niger, Nigeria, Palestine, Rwanda, Réunion,
-        Sao Tome and Principe, Senegal, Sierra Leone, Somalia, South Africa,
-        South Sudan, Sudan, Eswatini, Togo, Tunisia, Uganda, Tanzania, Western Sahara,
-        Zambia, Zimbabwe
+        Mozambique, Namibia, Niger, Nigeria, Rwanda, Réunion, Sao Tome and Principe,
+        Senegal, Sierra Leone, Somalia, South Africa, South Sudan, Sudan, Eswatini,
+        Togo, Tunisia, Uganda, Tanzania, Western Sahara, Zambia, Zimbabwe
       note: Depending on the spatial resolution, some models report Northern Africa
         as part of the Middle East region.
   - China+ (R10):
@@ -181,8 +180,8 @@
   - Middle East (R10):
       description: Middle East
       ar6: R10MIDDLE_EAST
-      countries: Bahrain, Iraq, Iran, Israel, Jordan, Kuwait, Lebanon, Oman, Qatar,
-        Saudi Arabia, Syria, United Arab Emirates, Yemen
+      countries: Bahrain, Iraq, Iran, Israel, Jordan, Kuwait, Lebanon, Oman, Palestine,
+        Qatar, Saudi Arabia, Syria, United Arab Emirates, Yemen
       note: Depending on the spatial resolution, some models report Northern Africa
         as part of the Middle East region.
   - North America (R10):
--- a/definitions/region/common.yaml
+++ b/definitions/region/common.yaml
@@ -194,12 +194,12 @@
       countries: Australia, Japan, New Caledonia, New Zealand, Samoa, Solomon Islands,
         Vanuatu
   - Reforming Economies (R10):
-      description: Former Soviet Union
+      description: Former Soviet Union countries not included in Europe
       ar6: R10REF_ECON
       countries: Armenia, Azerbaijan, Belarus, Georgia, Kazakhstan, Kyrgyzstan, Moldova,
         Russian Federation, Tajikistan, Turkmenistan, Ukraine, Uzbekistan
   - Rest of Asia (R10):
-      description: Other countries of Asia not included in China+ or India+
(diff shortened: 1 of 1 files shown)


byersiiasa (approved):
Looks fine and consistent- what about all the small countries that aren't listed... e.g. Curacao; Bonaire, Saint Eustatius and Saba

do you want a more comprehensive check for missing items?

danielhuppmann:
> what about all the small countries that aren't listed...

As far as I can see, the only missing countries have less than 1 million inhabitants, mostly small island states. So I would not add them until necessary, in favor of having a concise list.

danielhuppmann:
Per a discussion with @volker-krey, I added a clear indication that OECD membership is as of the year 1990, to explain why South Korea is not part of the OECD regions. Our preference is to not redefine the regions at this point.

christophbertram (approved):
This looks good. Only one observation: PAC_OECD looks to basically be Oceania plus Japan. The small oceanic islands are not in the OECD, but it probably does not harm to keep them in here. Perhaps adding this to description makes sure there is no misunderstanding. 
Regarding China+ and India+: The downside of the choice to keep relatively large countries in these groups is that this makes a larger difference between models with high regional resolution and those that don't, as the majority of models in the latter category have India and China as separate regions, but none of the other countries.

definitions/region/common.yaml
```diff
@@ -152,59 +153,59 @@
...
       countries: Canada, Guam, United States
   - Pacific OECD (R10):
-      description: Pacific OECD.
+      description: Pacific OECD (membership status 1990)
       ar6: R10PAC_OECD
       countries: Australia, Japan, New Caledonia, New Zealand, Samoa, Solomon Islands,
         Vanuatu
   - Reforming Economies (R10):
-      description: Former Soviet Union.
+      description: Former Soviet Union
```
byersiiasa, line 197: description: Former Soviet Union countries not included in R10Europe.

definitions/region/common.yaml
```diff
@@ -152,59 +153,59 @@
...
       ar6: R10REF_ECON
       countries: Armenia, Azerbaijan, Belarus, Georgia, Kazakhstan, Kyrgyzstan, Moldova,
         Russian Federation, Tajikistan, Turkmenistan, Ukraine, Uzbekistan
   - Rest of Asia (R10):
-      description: Other countries of Asia not included in China+ or India+.
+      description: Other countries of Asia not included in China+ or India+
```
byersiiasa, line 202: description: Other countries of Asia not included in China+, India+ or Reforming Economies

definitions/region/common.yaml
```diff
@@ -130,15 +131,15 @@
         Trinidad and Tobago, United States Virgin Islands, Uruguay, Venezuela
   - Other (R9):
       description: Rest of the World, used only if a match with the R9 regions can
-        otherwise not be achieved.
+        otherwise not be achieved
       navigate: R9OWO
 
 # The R10 definitions were introduced by the LIMITS project
 # funded by the EU's Seventh Framework Programme FP7/2007-2013
 # (grant agreement no. 282846)
```
byersiiasa, line 139: https://tntcat.iiasa.ac.at/LIMITSDB/static/download/LIMITS_10+1_regions.pdf

byersiiasa (commented):
LGTM

definitions/region/common.yaml
```diff
@@ -130,76 +131,80 @@
...
```
byersiiasa, line 139:
```suggestion
# (grant agreement no. 282846) https://tntcat.iiasa.ac.at/LIMITSDB/static/download/LIMITS_10+1_regions.pdf
```

definitions/region/common.yaml
```diff
@@ -130,76 +131,80 @@
...
       countries: Canada, Guam, United States
   - Pacific OECD (R10):
-      description: Pacific OECD.
+      description: Pacific OECD (membership status 1990)
       ar6: R10PAC_OECD
       countries: Australia, Japan, New Caledonia, New Zealand, Samoa, Solomon Islands,
         Vanuatu
   - Reforming Economies (R10):
-      description: Former Soviet Union.
+      description: Former Soviet Union
```
byersiiasa, line 197:
```suggestion
      description: Former Soviet Union countries not included in Europe
```

definitions/region/common.yaml
```diff
@@ -130,76 +131,80 @@
...
       ar6: R10REF_ECON
       countries: Armenia, Azerbaijan, Belarus, Georgia, Kazakhstan, Kyrgyzstan, Moldova,
         Russian Federation, Tajikistan, Turkmenistan, Ukraine, Uzbekistan
   - Rest of Asia (R10):
-      description: Other countries of Asia not included in China+ or India+.
+      description: Other countries of Asia not included in China+ or India+
```
byersiiasa, line 202:
```suggestion
      description: Other countries of Asia not included in China+, India+ or Reforming Economies
```

definitions/region/common.yaml
danielhuppmann: We don't know how much longer these databases will be accessible, so I'd rather not add a static link

danielhuppmann:
Re @christophbertram:

> PAC_OECD looks to basically be Oceania plus Japan. 

Thanks for that observation, the small-island states are actually duplicated in "Other Asia (R10)" - so I removed them from "Pacific OECD".

> China+ and India+:

The aim of the R10 regions, as far as I understood, was actually to have aggregrate-regions that are not directly identifiable as a specific country. This is to avoid too much focus on the country-level in IPCC assessments. Hence, I continue that logic (and make minimal changes to previous definitions).

PR #16, opened 2023-08-24
Add variables for electricity generation

danielhuppmann:
This PR introduces the variable tree for `Secondary Energy|Electricity` based on the ENGAGE and NAVIGATE projects.

A few observations:
- PV is separated into residential-rooftop and commercial-rooftop. Should we add a sub-category for utility-scale "Open Land"? And "Floating"?
- ENGAGE has a variable for electricity generation from ammonia, but NAVIGATE does not. Should we add it here?
- ENGAGE and NAVIGATE have variables for Curtailment, Storage, Storage Losses and Transmission Losses, but they seem inconsistent with the idea of an "electricity supply mix" of this category. How to proceed?

### DIFF
--- a/definitions/variable/energy/secondary-energy-electricity.yaml
+++ b/definitions/variable/energy/secondary-energy-electricity.yaml
@@ -13,7 +13,7 @@
 
 - Secondary Energy|Electricity|Non-Biomass Renewables:
     definition: Net electricity generation from hydro, wind, solar, geothermal, ocean,
-      and other renewable sources excluding biomass.
+      and other renewable sources excluding biomass
     unit: EJ/yr
 - Secondary Energy|Electricity|{Non-Biomass Renewables}:
     description: Net electricity generation from {Non-Biomass Renewables}
@@ -29,14 +29,6 @@
     definition: Net electricity generation from ammonia
     unit: EJ/yr
 
-- Secondary Energy|Electricity|Fossil:
-    description: Net electricity generation from fossil fuels including coal,
-      natural gas and conventional/unconventional oil
-    unit: EJ/yr
-- Secondary Energy|Electricity|Fossil|{CCS}:
-    description: Net electricity generation from fossil fuels including coal,
-      natural gas and conventional/unconventional oil {CCS}
-    unit: EJ/yr
 - Secondary Energy|Electricity|{Primary Fossil Fuel}:
     description: Net electricity generation from {Primary Fossil Fuel}
     unit: EJ/yr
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Per discussion with @volker-krey, the idea is to add sub-categories for PV and also add a variable for electricity generation from ammonia, but keep curtailment and storage (losses) out of the secondary-energy variable tree for the time being (and discuss further if it becomes relevant/necessary).

danielhuppmann:
Any comments, @IAMconsortium/common-definitions-coordination?

volker-krey (approved):
Thanks, Daniel. Looks good to me. After understanding the `{}` notation, I like the implementation to reduce repetition in the definitions.

definitions/variable/energy/secondary-energy-electricity.yaml
```diff
@@ -0,0 +1,41 @@
...
+- Secondary Energy|Electricity|Biomass|{CCS}:
+    definition: Net electricity generation from purpose-grown bioenergy crops, crop and
+      forestry residue, municipal solid waste bioenergy and traditional biomass {CCS}
+    unit: EJ/yr
+
+- Secondary Energy|Electricity|Non-Biomass Renewables:
+    definition: Net electricity generation from hydro, wind, solar, geothermal, ocean,
+      and other renewable sources excluding biomass
+    unit: EJ/yr
+- Secondary Energy|Electricity|{Non-Biomass Renewables}:
```
orichters, line 18: Do I understand that correctly that while `Secondary Energy|Electricity|Wind` is a subcategory of `Secondary Energy|Electricity|Non-Biomass Renewables`, it is not filed as `Secondary Energy|Electricity|Non-Biomass Renewables|Wind`. In the AR6 mapping file, we had sometimes both (for example [`Primary Energy|Non-Biomass Renewables|Wind`](https://github.com/pik-piam/piamInterfaces/blob/6209d8f52cf04a40cb5d1b2c67bae04e9ba3fe7c/inst/templates/mapping_template_AR6.csv#L1191) and [`Primary Energy|Wind`](https://github.com/pik-piam/piamInterfaces/blob/6209d8f52cf04a40cb5d1b2c67bae04e9ba3fe7c/inst/templates/mapping_template_AR6.csv#L1207) and that lead to some confusion, I think.

orichters (commented):
Looks good, just a short question…

definitions/variable/energy/secondary-energy-electricity.yaml
danielhuppmann: Correct, there was some confusion between trying to have a strictly hierarchical variable tree (Non-Biomass Renewables -> Wind) versus a more concise, readable approach (Wind).

Generally, the consensus in the community is (as far as I am aware) to prefer readability and have the additional information (which categories are part of e.g. Non-Biomass Renewables) either implicit or explicit (as an attribute of the definitions-yaml).

orichters: As you said that we will have explicit summation definitions and checks, that should cover these cases.

PR #17, opened 2023-08-29
Add MAGICC variables

phackstock:
As MAGICC becomes more integrated into a number of projects I figured it might be a good idea to add all MAGICC variables to this repo.

### DIFF
--- a/definitions/variable/climate/magicc.yaml
+++ b/definitions/variable/climate/magicc.yaml
@@ -1,0 +1,1376 @@
+- AR6 climate diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|10.0th Percentile:
+    unit: ppb
+- AR6 climate diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|16.7th Percentile:
+    unit: ppb
+- AR6 climate diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|17.0th Percentile:
+    unit: ppb
+- AR6 climate diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|25.0th Percentile:
+    unit: ppb
+- AR6 climate diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|33.0th Percentile:
+    unit: ppb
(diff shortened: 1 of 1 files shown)


definitions/variable/climate/magicc.yaml
```diff
@@ -0,0 +1,1376 @@
+- AR6 climate diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|10.0th Percentile:
```
danielhuppmann, line 1: For consistency with the variable-notation guidelines.
```suggestion
- AR6 Climate Diagnostics|Atmospheric Concentrations|CH4|MAGICCv7.5.3|10.0th Percentile:
```

danielhuppmann (commented):
Please add tag-lists for the percentiles and the species, see https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/emissions/tag_tier3_species.yaml (but you cannot directly re-use because of different units).

definitions/variable/climate/magicc.yaml
jkikstra: Is this PR still meant to be merged?

And if so, I suppose this capitalisation be changed directly in iiasa/climate-assessment?
I don't have any capacity for this right now (it's a small update, but updating the test-data is a bit of a pain, and just need to do a quick check to make sure it wouldn't break anything in notebooks or other scripts, and backward-compatibility).

phackstock: For now, I wouldn't say it's super urgent. MAGICC variables are currently added after the variable validation so this PR is good to have for sure, but not essential right now.

danielhuppmann:
FWIW, I think it might even be preferable to keep these variables in a separate repo directly attached to the climate-processing workflows - this way, the output of climate-processing is "together" with the expected variables. For validation, the variables could still be imported to a project-workflow repo (if necessary).

phackstock:
@danielhuppmann, I like the idea. We could and probably should encourage this for all future post-processing steps. Each step should come with it's own set of defined output variables.

jkikstra:
Okay sounds good, just let me know if/when you need anything from me.

PR #19, opened 2023-08-30
Add variables for secondary-energy production other than electricity

danielhuppmann:
This PR complements #16, adding variables for secondary-energy-carriers other than electricity.

Any comments, @IAMconsortium/common-definitions-coordination?


volker-krey (approved):
Thanks, Daniel. Looks good to me. After understanding the `{}` notation, I like the implementation to reduce repetition in the definitions.

definitions/variable/energy/secondary-energy-non-electricity.yaml
```diff
@@ -0,0 +1,115 @@
...
+- Secondary Energy|Solids:
+    definition: Total production of solid energy carriers, including coal, briquettes,
+      coke, and wood pellets
+    unit: EJ/yr
+- Secondary Energy|Solids|Biomass:
+    description: Production of solid secondary energy carriers from biomass
+      (e.g., commercial charcoal, wood chips, wood pellets).
+      Use of tradional bioenergy is excluded.
+    unit: EJ/yr
+- Secondary Energy|Solids|Fossil:
```
orichters, line 27: AR6 had `Secondary Energy|Solids|Coal`, but that seems to have changed in NAVIGATE. I'm just surprised that we keep `Secondary Energy|Gases|Coal`. Any reason for that?

definitions/variable/energy/secondary-energy-non-electricity.yaml
```diff
@@ -0,0 +1,115 @@
...
+    description: Production of solid secondary energy carriers from fossils
+      (e.g., briquettes, coke)
+    unit: EJ/yr
+
+- Secondary Energy|Gases:
+    definition: Total production of gaseous fuels, including natural gas and biogas,
+      not including hydrogen
+    unit: EJ/yr
+- Secondary Energy|Gases|Biomass:
```
orichters, line 36: Do we not consider CCS here because gas production does not allow for it?

orichters (commented):
Looks good, thanks, just two small questions.

definitions/variable/energy/secondary-energy-non-electricity.yaml
danielhuppmann: My intuition is that the gas (methane) still contains the carbon molecules, hence CCS would be accounted for at the point of combustion (e.g., in the power plant when generating electricity).

definitions/variable/energy/secondary-energy-non-electricity.yaml
danielhuppmann: Interesting observation, thank you! It seems that AR6 and ENGAGE used "Secondary Energy|Solids|Coal" whereas NAVIGATE used "Secondary Energy|Solids|Fossil".

Now that you mention it, I think "Secondary Energy|Solids|Coal" makes more sense, right?

Renato-Rodrigues: We had these discussions internally in the REMIND group and we decided, for our case, in favor of using the Fossils term using the following structure:

Secondary Energy|{se energy carrier}|{energy type input}|{pe energy carrier}

where:
se energy carrier = Electricity, Gases, Heat, Hydrogen, Liquids, Solids
energy type = Fossil, Biomass and Electricity
pe source = Oil, Coal, Gas, Biomass, Nuclear, Wind, Solar, Geothermal, Hydro, Ocean,...

Using fossils allow to have an direct aggregate to compare different formulations, like for example, one that split fossil waste from coal, from one that has everything aggregated. 

The same structure is also applied to our final energy variables reporting.

danielhuppmann: @Renato-Rodrigues, I think we should remain consistent with either one of the two previously existing variable names - changing all variables across the entire energy-variable tree will create a lot of headache for all modelling teams, so this should be avoided.

Renato-Rodrigues: My suggestion was in line of using the Fossil alternative, and I presented the REMIND scheme just as a means to justify way I suggested that.
I definitely agree with you that it makes no sense to make a huge change to the variables tree if it is not strictly necessary.

danielhuppmann: Ok, thanks for the clarification! Let's keep "Secondary Energy|Solids|Fossil" for consistency/simplicity, because all fossil-solids are coal (our some related type).

I revised the description of "Secondary Energy|Solids|Fossil" in line with "Secondary Energy|Gases|Gas", because both categories are basically just direct consumption of coal and natural gas (=fossil methane) with very limited transformation, only some purification and processing.

PR #33, opened 2023-11-08
Add capacity variables for secondary-energy-carriers

danielhuppmann:
This PR implements capacity-variables in parallel to the "Secondary Energy|..." variable structure.


christophbertram (commented):
Looks mostly good, but two questions:
Should the description for capacity values give an indication of whether gross or net capacity values are preferred?
And for non-electricity secondary energy, it might make sense to specify that this is about the "output" capacity, especially if we keep the unit as "GW" which for example for electrolysers is often the unit specifying the input capacity.

danielhuppmann:
This was implemented via #177 and #169

PR #37, opened 2023-11-08
Move energy use of carbon-removal to own section

danielhuppmann:
This PR does the following:
- move the variables for energy use for carbon removal to own section
- switch the ordering of the hierarchy from "fuel - technology" to "technology - fuel" for consistency with other "Final Energy" variables

Open questions: are there other carbon-removal technologies that should be included for reporting of their energy use?

Any comments @IAMconsortium/common-definitions-coordination @gidden?

### DIFF
--- a/definitions/variable/energy/final-energy.yaml
+++ b/definitions/variable/energy/final-energy.yaml
@@ -90,6 +90,40 @@
     description: Final energy consumption of non-renewable waste by the industrial sector
     unit: EJ/yr
 
+# energy use for carbon removal technologies
+- Final Energy|Carbon Removal:
+    description: Total energy use for carbon removal
+    unit: EJ/yr
+- Final Energy|Carbon Removal|Electricity:
+    description: Electricity use for carbon removal
+    unit: EJ/yr
+    notes: See `Secondary Energy|Electricity|...` for the power generation mix
+- Final Energy|Carbon Removal|{Secondary Fuel}:
+    description: Use of {Secondary Fuel} for carbon removal
+    unit: EJ/yr
+    notes: See `Secondary Energy|{Secondary Fuel}|...` for the sources of the energy carrier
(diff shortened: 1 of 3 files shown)


gidden:
In the GENIE project explorer, we used "Greenhouse Gas Removal" and included a number of variables (searchable by "Direct Air Capture"): https://data.ece.iiasa.ac.at/genie/#/docs

To be safe, I would include all CDR methods, as most if not all include an energy component to some degree (DACCS, EW, biochar, etc.).

danielhuppmann:
Thanks @gidden! Not sure how many models have non-CO2 removal technologies at this point explicitly modelled, so I'd prefer to follow the ENGAGE/NAVIGATE template for the time being. We can always add more variables later.

About technologies, GENIE also only reported energy use of Direct Air Capture, so can you suggest a specific list of technologies that should be included?

strefler:
At least Enhanced Weathering and Ocean Alkalinity Enhancement should be included, but probably also biochar and materials.

gidden:
Hey @danielhuppmann - here are the list of CDR methods we 

> DACCU/DACCS
> Ocean Alkalinization
> Enhanced Weathering
> BECCU/BECCS
> Aforestation/Reforestation
> Soil Carbon Sequestration
> Biochar
> Peatland Restoration
> Wetland Restoration
> Agroforestry

In particular, I think DAC, OAE, EW, and Biochar will have nontrivial energy usage. We could also have an "Other" catchall if we want to consider other methods. The additional energy for biochar can definitely be reported, as it requires pyrolysis.

danielhuppmann:
Thanks @strefler and @gidden!

Anyone willing to approve this PR? Or other items that should be included right away...?

gidden (approved):
Lgtm thanks @danielhuppmann !

PR #38, opened 2023-11-09
Add multiple versions of carbon-price variables

danielhuppmann:
This PR adds the carbon-price variables (with different forms of aggregation) as used in the ENGAGE project.

fyi @IAMconsortium/common-definitions-coordination

### DIFF
--- a/definitions/variable/socio-economic/carbon-price.yaml
+++ b/definitions/variable/socio-economic/carbon-price.yaml
@@ -1,0 +1,26 @@
+- Price|Carbon:
+    description: Price of carbon (as reported by the model)
+    unit: USD_2010/t CO2
+    region-aggregation:
+      - Price|Carbon (mean):
+          method: mean
+      - Price|Carbon (weighted by Emissions|CO2):
+          weight: Emissions|CO2
+          drop_negative_weights: false
+      - Price|Carbon (weighted by Final Energy):
+          weight: Final Energy
+    check-aggregate: false
+- Price|Carbon (mean):
+    description: Price of carbon (computed as unweighted mean across regions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+- Price|Carbon (weighted by Emissions|CO2):
+    description: Price of carbon (weighted by regional CO2 emissions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+    note: Regional CO2 emissions can turn negative at different points in time, which
+      can result in counter-intuitive timeseries when used as weights for averages
+- Price|Carbon (weighted by Final Energy):
+    description: Price of carbon (weighted by final energy consumption across regions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true


orichters:
In NGFS, we also have sectoral-differentiated carbon prices because we have sector-specific policies.
Worth adding something like this?
```
Price|Carbon|Demand|Industry
Price|Carbon|Demand|Residential and Commercial
Price|Carbon|Demand|Transportation
```

danielhuppmann:
> In NGFS, we also have sectoral-differentiated carbon prices because we have sector-specific policies. 

No objections from me, but looking at the NGFS variable, I notice two issues:
1. these variables do not have a description
2. these variables are all weighted by "Final Energy" (not the emissions or energy use of that sector) - is that intended?

I suggest do a follow-up PR (and discuss the two issues there) once this PR is merged (after one week unless there are any objections).

orichters:
We weigh by Final Energy because once you go to negative emissions in some sectors, weighing by emissions creates really strange effects such as massive spikes (dividing by ~ zero) and negative carbon prices. Fine to discuss that later.

orichters:
Question: Is it really a good idea to have variables that include a part within parantheses? I'm asking because many of our scripts use strings of the format "Variable name (Unit)" and I'm a bit concerned that may mess up some of our scripts :)

I couldn't find these variable names in the Elevate template I have, anyway (from May 26, 2023)

danielhuppmann:
Right, this was added in the openENTRANCE project via https://github.com/openENTRANCE/openentrance/pull/154 per suggestion by @Renato-Rodrigues (after discussion with @robertpietzcker, if I recall correctly). The idea is to have an easy way to compare different computations for carbon price given the caveats of using negative weights.

If you think that this is too much of a distraction, I can also remove it and we start a separate PR and discuss there...

orichters:
No, that is fine. If Renato proposed it, I have no doubt it is a good idea. And with the parantheses: As this basically affects only postprocessing scripts (we don't need to name it like that in REMIND), no objection to that.

definitions/variable/socio-economic/carbon-price.yaml
```diff
@@ -0,0 +1,26 @@
...
+    check-aggregate: false
+- Price|Carbon (mean):
+    description: Price of carbon (computed as unweighted mean across regions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+- Price|Carbon (weighted by Emissions|CO2):
+    description: Price of carbon (weighted by regional CO2 emissions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+    note: Regional CO2 emissions can turn negative at different points in time, which
```
orichters, line 21: I wonder whether aggregating by `Gross Emissions|CO2` would be preferable, avoiding this really counter-intuitive and confusing issue of negative or undefined (division by 0) carbon prices

danielhuppmann: Not sure about that, this solution would also deviate from the "standard" solution where the price is weighted by net emissions.

robertpietzcker: There is no "perfect" solution. In my view, the version "weighted by final energy" would be the best default setting, as it represents the angle "how large is the energy system for which a carbon price is relevant" (this could also be "primary", but I think final is a bit better), but if one takes a different angle, other aggregations can be more fitting. 

If there is a specific angle for which weighting by "gross emissions" would be better, it could be added as well, right?

FlorianLeblancDr: Gross and Net emissions are both fine.
Having negative mean pricing should be rare: in my understanding, this would happened if regions without NET have higher carbon prices than in regions with NET, which is really counter-intuitive and is more a modelling issue.

In the past Gross emissions were used (previous papers and the IPCC) to compute the net present value of the global carbon price in order to cover negative prices, but again I think at that time the issue was the use of a discount rate not compatible with some model results, rather than using "Gross" over "Net" emissions.

Anyway those should be moved to the "macro" list now.

FlorianLeblancDr: Thank for this numerical example @robertpietzcker 
I agree then.

With the caveat: carbon price trigger energy efficiency, which in this case decrease the weight applied.

danielhuppmann:
Closing in favor of #289

Issue #36, opened 2023-11-08
Variable structure for emissions and energy use of bunker fuels


danielhuppmann:
When merging the ENGAGE and NAVIGATE variable templates, I noticed an issue with the treatment of bunker fuels...

# The problem

Emissions and energy use for international shipping and aviation (aka "bunkers") are not consistently counted in (national) emission inventories and statistics, instead treated as memo items, etc.

# Solution adopted in ENGAGE and NAVIGATE

...and initially transferred into this new variable template:

1. Have international bunkers as a separate sector outside of Transportation, e.g., 
    - Final Energy|Bunkers|International Aviation
    - Final Energy|Bunkers|International Shipping
    - ...

   in contrast to domestic aviation and shipping
    - Final Energy|Transportation|Domestic Aviation
    - Final Energy|Transportation|Domestic Shipping

    and sub-categories, parallel for emissions...

2. There are several "memo items" for comparability with relevant statistics
    - Emissions|CO2 (w/o bunkers)
    - Emissions|CO2|Energy|Demand|Transportation (w/ bunkers)
    - Final Energy (w/o bunkers)
    - Final Energy|Transportation (w/ bunkers)

# Problems going forward

1. As a consequence, energy use from green fuels (e.g., battery, hydrogen planes) is accounted within the bunkers category, even though they are not "bunker fuels"...
2. The "memo items" are once "with bunkers", once "without bunkers", which may be confusing...

# Alternative proposal

1. Move international transportation energy use and emissions back into the "Transportation" variable tree
2. Make memo items consistently as "w/o bunkers"

To be explicit, the proposal is:

- Final Energy
- Final Energy (w/o bunkers) [Memo Item]
- Final Energy|Transportation
- Final Energy|Transportation (w/o bunkers) [Memo Item]
- Final Energy|Transportation|Domestic Aviation
- Final Energy|Transportation|Domestic Shipping
- Final Energy|Transportation|International Aviation
- Final Energy|Transportation|International Shipping
- Final Energy|...

and sub-categories, parallel for emissions...

And it should be clarified whether the memo items includes non-fossil energy use of international shipping and aviation...

Any comments @IAMconsortium/common-definitions-coordination?

strefler:
The reasons for the w/ or w/o bunkers is historic: We wanted to avoid keeping the same variable name, but changing the definition either from w/ to w/o bunkers or the other way around. So in order to keep backwards compatibility with older reporting templates, I suggest to keep this.
[...]

PR #40, opened 2023-11-13
Change "Bunkers" sector to mean international aviation and shipping

danielhuppmann:
This PR responds to a comment by @strefler in #36. It (proposes to) change the meaning of the Bunkers sector include all international aviation and shipping to avoid an inconsistency between oil and green fuels in international transportation.

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -42,13 +42,13 @@
 - Emissions|{Tier-3 Species}|Energy and Industrial Processes:
     description: Emissions of {Tier-3 Species} from energy use in supply and demand sectors
       (IPCC category 1A, 1B) and from industrial processes (IPCC categories 2A, B, C, E),
-      including emissions from international bunkers, net of negative emissions from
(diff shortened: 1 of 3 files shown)


danielhuppmann:
Per a discussion with @volker-krey today, I updated all descriptions related to transportation and bunkers such that
- transportation clearly excludes international shipping and aviation
- "bunkers" refers to the International aviation and shipping sector, not just the fuel, so electricity consumption by international aviation is counted as "Final Energy|Bunkers|Electricity"

volker-krey:
Thanks. The proposed revisions look good to me. Given that many IAMs won't be able to distinguish between domestic and international aviation, I would propose that as a fallback, we ask teams to report all aviation under domestic aviation. Looking at IEA statistics, in 2021 roughly 50% of aviation was international and 50% domestic. With IAMs aggregating many countries into regions, this would lead to "within region" aviation becoming a larger share. For example, in the US (which is maybe the size of an average IAM region), about two thirds of aviation is domestic and one third international. For international shipping the situation is different with globally three quarters being international and one quarter domestic navigation. So here in doubt reporting international would make sense.

Issue #34, opened 2023-11-08
Change energy-carrier structure for heat generation fuel mix


danielhuppmann:
While working on #33, I noticed that the variable hierarchy for heat generation in the ENGAGE and NAVIGATE templates does not account for any "green" fuels like biofuels, biomethane or hydrogen.

The current structure (omitting the CCS-subcategories) is:

- Secondary Energy\|Heat\|Biomass
- Secondary Energy\|Heat\|Electricity
- Secondary Energy\|Heat\|Fossil
- Secondary Energy\|Heat\|Coal
- Secondary Energy\|Heat\|Gas
- Secondary Energy\|Heat\|Oil
- Secondary Energy\|Heat\|Geothermal
- Secondary Energy\|Heat\|Nuclear
- Secondary Energy\|Heat\|Solar
- Secondary Energy\|Heat\|Other

where for "Gas" is not clear whether this only refers to fossil methane. (We usually use "Gases" for the aggregate of fossil and renewable forms of methane).

I suggest to apply the "Final Energy"-convention to the heat generation fuel mix and use "Liquids", "Solids" and "Gases" instead of "Oil", "Coal" and "Gas" (with possible sub-categories to distinguish between fossil and renewable types of the fuel).

Any comments @IAMconsortium/common-definitions-coordination?

Renato-Rodrigues:
I second the suggestion of using the final energy convention and I would like to also suggest adding a further split of all of those sub-categories, in final and secondary energy, into `Fossil`, `Biomass` and `Electricity (e-fuels)`, as it is currently being done at the ECEMF template:

e.g.
```
Secondary Energy|Heat|Liquids|Fossil, Biomass or Electricity
Secondary Energy|Heat|Gases|Fossil, Biomass or Electricity
Secondary Energy|Heat|Solids|Fossil or Biomass
Secondary Energy|Heat|Electricity
Secondary Energy|Heat|Geothermal
Secondary Energy|Heat|Nuclear
Secondary Energy|Heat|Solar
Secondary Energy|Heat|Other
```

danielhuppmann:
In this variable template so far, I have not included that type of information. Copying (and slightly adapting) a blurb from my email to the PRISMA consortium on August 31, 2023:

[...]

PR #41, opened 2023-11-20
Clarify meaning of biomass in secondary-energy variables

danielhuppmann:
Per a comment by @strefler in https://github.com/IAMconsortium/common-definitions/issues/34#issuecomment-1806065218, this PR changes the description of secondary-energy variables to clarify that biomass includes biogas


danielhuppmann:
Merging ahead of the IAMC SWG meeting

PR #53, opened 2024-02-13
add SDG template from SHAPE

bs538:
First draft of SDG template, copied as is from SHAPE project SDG template.

Note: this SDG template was used together with the NAVIGATE template, therefore SDG-relevant variables already contained in the NAVIGATE template are not listed again here. Once the other sub-templates are developed a bit further, we need a round of checking them and cross-tagging SDG-relevant variables.

PR #54, opened 2024-02-13
Add GDP variables

danielhuppmann:
This PR adds GDP variables in response to #53 using the 2010-USD as unit - however, the latest SSP-release (see http://data.ece.iiasa.ac.at/ssp/) used 2017-USD as unit for GDP projections.

Any suggestions/preferences on which one to use?
@IAMconsortium/common-definitions-coordination @IAMconsortium/common-definitions-macro-economy

### DIFF
--- a/definitions/variable/macro-economy/gdp.yaml
+++ b/definitions/variable/macro-economy/gdp.yaml
@@ -1,0 +1,6 @@
+- GDP|MER:
+    definition: GDP converted to USD at market exchange rate (MER)
+    unit: billion USD_2010/yr
+- GDP|PPP:
+    definition: GDP converted to USD using purchasing power parity (PPP)
+    unit: billion USD_2010/yr


FlorianLeblancDr:
this issue applies to many economic variable @IAMconsortium/common-definitions-macro-economy 

I would suggest to stick to $2010, for two reasons:

- practice will evolve on current projects (PRISMA, SPARCLE, ACCRUE, etc.)
- The base year of the SSP database is not systematically taken in models. A few example: IMACLIM and E3ME uses $2014 from GTAP-10. GEM-E3 uses $2017 from GTAP-11. And teams simply use a GDP deflator to convert economic values in $2010

It is also better to set the reference year for constant economic value in only one place, using a tag.
We are working on such a suggestion with the macro-economic subgroup (adding a Currency-Tag)

This will allow specific templates derived from the common one to change this reference year easily for all variables.

lolow:
I agree that monetary values should be consistent across the same database. Implementing a currency tag could be a solution. However, it should not be used in the same way as other tags, rather as a placeholder for a monetary unit to be defined at database level.

lolow (approved):

danielhuppmann:
Thank you for this quick and productive discussion!

PR #61, opened 2024-02-15
Guard against adding deprecated legacy variables

danielhuppmann:
This PR adds a test and GitHub Actions workflow to guard against mistakenly adding variables that were marked as being superseded by new variables.

The way that legacy variables are marked in the common-definitions repository is by having an attribute with the legacy project name, currently "navigate" and "engage".

```yaml
- Final Energy|Carbon Removal|{Carbon Removal Option}|Electricity:
    description: Electricity use for carbon removal by {Carbon Removal Option}
    unit: EJ/yr
    notes: See `Secondary Energy|Electricity|...` for the power generation mix
    navigate: Final Energy|Carbon Removal|Electricity|{Carbon Removal Option}
```

The test workflow checks that all such attributes are not duplicates of variables in the VariableCodeList.

### DIFF
--- a/.github/workflows/legacy.yml
+++ b/.github/workflows/legacy.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Set up Python 3.11
       uses: actions/setup-python@v4
--- a/.github/workflows/legacy.yml
+++ b/.github/workflows/legacy.yml
@@ -1,0 +1,32 @@
+# This workflow validates that (new) variables are not referenced as deprecated legacy variables
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Validate the project
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  project-validation:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v4
+      with:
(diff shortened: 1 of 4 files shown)


danielhuppmann:
The validation can be seen "in action" [here](https://github.com/IAMconsortium/common-definitions/actions/runs/7918318624/job/21616446807?pr=61#step:6:77)

requirements.txt
phackstock, line 1: I don't think you need a dedicated requirements file. I'd just put nomenclature-iamc into the workflow file like in legacy-definitions. Like you're doing it already with pytest.
```
pip install nomenclature-iamc pytest
```

.github/workflows/legacy.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  legacy-validation:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
```
phackstock, line 18: There's now version 4 of checkout
```suggestion
    - uses: actions/checkout@v4
```

.github/workflows/legacy.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.11
+
+    - name: Install requirements
+      run: pip install -r requirements.txt
```
phackstock, line 26: I think you can remove this and just install nomenclature along with pytest below.

phackstock (approved):
Looks good to me, just a few small comments but already good to merge.

danielhuppmann:
Thanks @phackstock - I prefer to keep the requirements.txt file for now because this is an easy way for new users to quickly see what is required to run this project. We can add it to the Readme instead in a follow-up PR.

PR #66, opened 2024-03-02
Add details to emissions variables related to industrial processes

danielhuppmann:
This PR implements some suggestions by the @IAMconsortium/common-definitions-emissions group related to emissions from industrial processes.


danielhuppmann:
Merging per approval by @jayfuhrman (see email March 8, 2024).

PR #67, opened 2024-03-02
Add variables for residual emissions and CDR

danielhuppmann:
This PR adds variables for residual emissions of CO2 and for carbon-dioxide removal - however, this seems to be a confict with the variables "Carbon Removal" and "Carbon Removal|\<type\>" used in the NAVIGATE project.

Please advise @IAMconsortium/common-definitions-emissions

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -11,6 +11,23 @@
       and other land use (AFOLU)
     unit: Mt CO2/yr
 
+- Emissions|CO2|Residual Emissions:
+    description: Gross emissions of carbon dioxide (CO2) from energy use in supply and
+      demand sectors (IPCC category 1A, 1B) and from industrial processes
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Looking at this PR again, I find it strange that **Emissions|CO2|Carbon Dioxide Removal** should be reported as a positive number - this goes against the logic of other **Emissions|CO2|...** variables where negative indicators removal.

strefler:
I may have missed some of the discussion, but why are we adding those variables? Shouldn't Gross Emissions|CO2 be exactly the same as the residual emissions? And Carbon Removal the same as Emissions|CO2|CDR, except with a positive sign? If we decide to keep these, I would agree that Emissions|CO2|CDR should be negative

danielhuppmann:
I guess that the initial request to add these variables is now no longer relevant - should this PR be closed?

strefler:
Yes, that is my understanding

danielhuppmann:
Closing as superseded by the subsequent discussions

PR #68, opened 2024-03-02
Refactor final-energy-variables for CO2 removal to carbon management

danielhuppmann:
Noting that the variables added in #37 are inconsistent because Direct-Air-Capture is not really a removal-technology. Therefore, this PR refactors the related variables to the broader term "Carbon Management".


jayfuhrman:
Agree with this revision to be more generic in terminology.  Not all captured carbon is necessarily removed from the atmosphere and instead reflects avoided emissions if the source is fossil or industrial rather than bio or direct air capture.  Also even some hopefully small but non-zero fraction direct air capture CO2 is likely to leak and this structure enables us to reflect this.

danielhuppmann:
Added reference to legacy variables from NAVIGATE after renaming.
Merging per approval by @jayfuhrman (also see email March 8, 2024).

PR #71, opened 2024-03-03
add Techno-economic

beijingzyl:


danielhuppmann:
Thank you @beijingzyl for opening this PR. This seems to be a straight copy from the NAVIGATE template, correct?

Can you please let me know whether you have had any discussions in the subgroup about possible improvements/extensions to the variables? For example, we aim to ensure easy readability of the variables and avoid abbreviations, so renaming "OM Cost" to "Operation and Maintenance Cost" might be worth considering. Also, there is a sub-category "Fixed" in the OM cost variables but there is specification of the "non-fixed" component...

beijingzyl:
@danielhuppmann Thank you for your feedback! This PR is from the AR6 template. We've made modifications, such as changing "capital cost" to "techno-economic" to better reflect its scope. We appreciate your suggestions on naming conventions and will consider them.

Regarding the "Fixed" in the OM cost variables, it indeed refers to fixed O&M costs. We noted there's no mention of a "non-fixed" counterpart, and it appears not to be a concept in AR6 either.

As for the validating error, it seems to be related to units after reviewing the error message, but I couldn't identify the exact issue. It's possible that the error arises from a mismatch between the expected and provided units, or there might have been incorrect unit conversions.

danielhuppmann:
Thank you, @beijingzyl.

The tests for your PR are currently failing because the yaml file is not in the right format - it should be a list (each item starting with `-`, not a yaml-dictionary).

A better starting point for your work can be found [here](https://github.com/IAMconsortium/legacy-definitions/blob/main/definitions/variable/variable.yaml), which is already in the right format. Also, please use all-lower-case names for files.

For units, please follow the new convention "USD_2010" instead of "US$2010" so that we can use the [iam-units](https://github.com/iamconsortium/units) package for automated handling of unit validation and conversion. All other variables in this repo already follow this convention.

About the fixed O&M costs variables: the purpose of developing the common-definitions repository as a community activity is to review which variables make sense or should be revised. We set up a variable naming guidance [here](https://docs.ece.iiasa.ac.at/iamc/variable.html#variable-naming-conventions) and the variables "OM Cost|Fixed|..." use an abbreviation and establish an un-used hierarchy level. Please discuss with the @IAMconsortium/common-definitions-techno-economic group on a better variable structure.

ouyang363 (approved):
initial commit, consistent with AR6

danielhuppmann:
Thank you @beijingzyl and @ouyang363 for implementing part of the requested changes.

However, you did not respond to my comment about improving the variable name for operation-and-maintenance costs:
> About the fixed O&M costs variables: the purpose of developing the common-definitions repository as a community activity is to review which variables make sense or should be revised. We set up a variable naming guidance [here](https://docs.ece.iiasa.ac.at/iamc/variable.html#variable-naming-conventions) and the variables "OM Cost|Fixed|..." use an abbreviation and establish an un-used hierarchy level. Please discuss with the @IAMconsortium/common-definitions-techno-economic group on a better variable structure.

danielhuppmann:
See https://github.com/beijingzyl/Techno-economic/pull/1 for a simplification and one specific question related to this PR.

Note that https://github.com/beijingzyl/Techno-economic/pull/1 is directed into the branch for this PR, so you first have to discuss and merge that other PR, then continue this work.

danielhuppmann:
Closing in favor of #201

PR #76, opened 2024-03-21
add CCS to cmanagement definitions

gidden:
FYI @strefler @jayfuhrman @danielhuppmann 

Are we ok with just one FE CCS variable? There is a bit of overlap with DAC here. I would say energy for purely the capture of DAC go into the DAC variable, but transportation and storage energy (if modeled) go into CCS.

### DIFF
--- a/definitions/variable/emissions/tag_carbon-management.yaml
+++ b/definitions/variable/emissions/tag_carbon-management.yaml
@@ -7,6 +7,8 @@
         description: production of biochar using pyrolysis
     - Direct Air Capture:
         description: direct air capture
+    - Carbon Capture and Storage:
+        description: the capture from point sources, including distribution and ultimate storage
     - Enhanced Weathering:
         description: terrestrial application of alkalinity to increase CO2 uptake on land
     - Direct Ocean Capture:
--- a/definitions/variable/emissions/tag_carbon-management.yaml
+++ b/definitions/variable/emissions/tag_carbon-management.yaml
@@ -3,12 +3,13 @@
 
 # This list should be a subset of carbon-removal options in definitions/variable/emissions/carbon-removal.yaml
 - Carbon Management Technology:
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Thanks @gidden!

I think there are three options.
1. DAC - direct air capture only, no sequestration
2. CC - point-source capture only, no sequestion
3. Carbon Transport and Sequestration

Or include 3 always with 2 (which is your initial proposal, if I understand it correctly) - then you should also update the DAC description

Or split 3 between 1 and 2 based on shares - and you should also clarify the DAC description accordingly.

gidden:
@jayfuhrman re-pinging you on this. any further thoughts on energy variables?

strefler:
Agree with Daniel. From my point of view it would make sense to distinguish carbon capture (either from a point source, or from the atmosphere, or the ocean) from carbon transport and storage, also in terms of energy requirements

danielhuppmann:
Is it still relevant to add a variable for final energy use of carbon-capture at point-sources and/or injection into geological storage? If yes, please rebase this PR or start a new one.

Currently, we have variables **Final Energy|Carbon Management|...** for the technologies defined at https://github.com/IAMconsortium/common-definitions/blob/34d84a60a2d46f9b4d7b71592f0f85222266d943/definitions/variable/emissions/tag_carbon-management.yaml#L5

jayfuhrman:
@danielhuppmann I think it makes sense to include point source CCS here, and to not refer to the subset of carbon-removal options in the tag_carbon_management.yaml file.  However unfortunately I don't appear to have write access so could you assist with the rebase?

danielhuppmann:
Rebased - is this good to be merged?

strefler:
Yes, works for me. We may still want to think about separating the energy requirements for only the transport and storage part, but right now we would not be able to report that separately, so fine with me!

PR #79, opened 2024-03-21
Update tag_carbon-management.yaml

jayfuhrman:
Add variables for Carbon Capture with levels of disaggregation based on the fate of the captured CO2 (i.e., geologic storage, utilization, other storage)

@strefler @fschreyer @gidden @Renato-Rodrigues @ouyang363, others please comment as needed

### DIFF
--- a/definitions/variable/emissions/tag_carbon-management.yaml
+++ b/definitions/variable/emissions/tag_carbon-management.yaml
@@ -1,38 +1,3 @@
-
-- Carbon Capture:
-    description: Captured carbon dioxide (CO2) from energy use in supply and demand sectors (IPCC category 1A, 1B) and from industrial processes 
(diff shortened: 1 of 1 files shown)


fschreyer:
Looks good to me, overall. Not sure whether we already raised this somewhere but would it be worth to differentiate carbon capture into sources and sinks? Like

```
Carbon Capture|Source|Industry|Biomass
...
Carbon Capture|Sinks|Geological Storage
...
```
Do we have variables for the different sources of carbon capture already?

gidden:
Just an FYI that I think we should discuss this an #80  in our upcoming call. I guess it should probably be 1 PR for both?

danielhuppmann:
Closing in favor of #83

PR #82, opened 2024-03-22
Add CDR technologies to final-energy reporting

jayfuhrman:


danielhuppmann:
@gidden @jayfuhrman, are you sure that these additional technologies should be added to the final-energy reporting? In other words, are their energy-requirements so large as to be relevant?

danielhuppmann:
Thanks for the answer - will merge this PR if @gidden approves.

gidden:
LGTM, thanks!

Issue #55, opened 2024-02-13
Units/methods/operations-specifications in the variable name


danielhuppmann:
In previous projects, the variable names sometimes included specifications of units, methods or operations, e.g.
- Final Energy|Electricity|Share - to distinguish from the absolute amount of electricity use
- Trade|Emissions Allowances|Volume - to distinguish from value of trade in monetary terms
- Price|Agriculture|Livestock|Index - to make clear that these are indexed values

This is an abuse of the hierarchical structure and it makes variable-names difficult to read (for humans) and complicates automated processing (for machines).

I therefore suggest to change the convention to include a unit/method/operations identifier in brackets instead of appending it as a sub-sub-[sub-]category (and remove the identifier altogether where it is not necessary).

According to this proposal, the variables should then read as follows:
- Final Energy|Electricity (Share)
- Trade|Emissions Allowances (Volume)
- Price|Agriculture|Livestock (Index)

@IAMconsortium/common-definitions-coordination

FlorianLeblancDr:
I like your proposal, at least this is better than using the hierarchical structure.
A few comments :

**Volume**
Should we consider that without brackets, by default the variable should be understood as volumes?
Most variables are currently in volume in the database, and this will avoid renaming if later "Index" or "Share" is introduced.

**Index**
Definitely useful when models do not agree on the unit (for example the activity of freight is measured in ton/km or in pseudo-$ depending of the model).

**Share**
I wonder if having "Shares" variables is not misleading, as it let each modelling team decide what is the denominator.
Isn't it equivalent to request the variable you want to compute shares?

For example I have the employment variable. Instead of requesting employment as a share of something (something being specified in the definition), I could only request (for example) total active population and let the share be computed elsewhere?

Example:
`- Unemployment:
    definition: Number of unemployed inhabitants (based on ILO classification)
    unit: million
- Unemployment|Rate:
    definition: Fraction of unemployed inhabitants (based on ILO classification)
    unit: '%'
`
To be replaced by:
`- Unemployment:
    definition: Number of unemployed inhabitants (based on ILO classification)
    unit: million
- Population|Active:
    definition: Active population.
    unit: '%'
`

May be this last point on "Share' deserve a specific issue.

danielhuppmann:
Thanks @FlorianLeblancDr for the response.

1. Yes, we should definitely review whether these identifiers are relevant and we should remove them where possible.
[...]

Issue #46, opened 2023-12-06
Backwards-compatibility of variable names


khaeru:
At the SWG meeting on 2023-12-06, Masa Sugiyama and others raised the idea of how to support backward-compatibility if it becomes necessary to change a variable name.

This issue is to discuss/collect ideas.

khaeru:
My suggestion:
- In the NAVIGATE project, we made use of the fact that the nomenclature package (which is used with this repo) tolerates (or reads and stores?) extra entries in code lists. For example, we had:
  ```yaml
  - NAV_Dem-20C-all_u:
      navigate_task: T3.5
      navigate_climate_policy: 20C
      navigate_T35_policy: act+ele+tec
  ```
  In this, `navigate_T35_policy` is like `description`, `units`, or other attributes.
- This is analogous to/imitates the SDMX concept of an Annotation.
- We should simply **specify a common annotation ID** that would contain 1 or a list of older/superseded/alias variable names. For instance:
  ```yaml
  - Final Energy|Foo|Bar:
      iamc-variable-superseded: |
        Final Energy|Bar|Foo
        Final Energy|Foo Bar
  ```
  It could be `iamc-variable-synonym`, `iamc-variable-old`, or anything—I don't have any strong preference here.
- Code that needs to handle older data could then access these annotations for info on the correspondence of old and current names, for instance to construct a "mapping" or "table", perform replacement, or whatever makes sense in a particular implementation.

<details>
<summary>A minimal working example (MWE) using SDMX:</summary>

```py
import sdmx
import sdmx.model.v21 as m

# Create a Code whose ID is a current variable name
c = m.Code(id="Final Energy|Foo|Bar")

# Create an annotation containing old/superseded variable names
ann = m.Annotation(
    id="iamc-variable-old",
    text="\n".join(
        ["Final Energy|Bar|Foo", "Final Energy|Foo Bar"],
    )
)
c.annotations.append(ann)

# Write to file
cl = m.Codelist(id="VARIABLE", name="IAMC variable name")
cl.append(c)
msg = sdmx.message.StructureMessage()
msg.add(cl)
with open("example.xml", "wb") as f:
    f.write(sdmx.to_xml(msg, pretty_print=True))
```

This gives output like:
```xml
…
  <str:Code id="Final Energy|Foo|Bar">
    <com:Annotations>
      <com:Annotation id="iamc-variable-old">
        <com:AnnotationText xml:lang="en">Final Energy|Bar|Foo
[...]

PR #89, opened 2024-04-08
Guard against methods-terms as sub-categories

danielhuppmann:
In #55, we agreed to use square brackets to identify methods or operations, e.g. `... [Share]` instead of the previously used subcategories like `...|Share`.

This PR adds a test to guard against adding subcategory-methods-variables in the future and renames one variable from the SDG domain.

fyi @IAMconsortium/common-definitions-sdg-indicators

### DIFF
--- a/definitions/variable/sdg-indicators/sdg.yaml
+++ b/definitions/variable/sdg-indicators/sdg.yaml
@@ -98,12 +98,12 @@
     sdg: 2
     unit: kcal/cap/day
     weight: Population
-- Price|Agriculture|Livestock|Index:
+- Price|Agriculture|Livestock [Index]:
     description: Weighted average price index of livestock
     sdg: 2
     unit: Index (2020 = 1)
     skip-region-aggregation: true
-#     weight: Agricultural Production|Non-Energy
+    shape: Price|Agriculture|Livestock|Index
 - Health|Premature Deaths|PM2.5:
     description: Number of premature deaths associated with increased health risks from
       exposure to air pollution of fine particles with a diameter of 2.5 μm or less (PM2.5)
--- a/tests/test_reserved_terms.py
+++ b/tests/test_reserved_terms.py
@@ -8,9 +8,9 @@ def test_variable_ops_as_square_brackets():
     dsd = DataStructureDefinition("definitions/", dimensions=["variable"])
 
     error = []
-    for variable, attrs in dsd.variable.items():
+    for variable in dsd.variable:
         if reserved_terms := [r for r in ["Index", "Share"] if "|" + r in variable]:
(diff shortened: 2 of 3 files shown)


definitions/variable/sdg-indicators/sdg.yaml
```diff
@@ -98,12 +98,12 @@
...
     unit: kcal/cap/day
     weight: Population
-- Price|Agriculture|Livestock|Index:
+- Price|Agriculture|Livestock [Index]:
     description: Weighted average price index of livestock
     sdg: 2
     unit: Index (2020 = 1)
     skip-region-aggregation: true
-#     weight: Agricultural Production|Non-Energy
+    shape: Price|Agriculture|Livestock|Index
```
phackstock, line 106: Maybe I missed this use, I don't know what the `shape` attribute is.

tests/test_reserved_terms.py
```diff
@@ -0,0 +1,17 @@
+from nomenclature import DataStructureDefinition
+
+
+def test_variable_ops_as_square_brackets():
```
phackstock, line 4: Should this not go into nomenclature as part of `VariableCodeList`? I think we would want to enforce these standards also outside of common definitions going forward. Maybe with the option to suppress the error for legacy variable names.

tests/test_reserved_terms.py
```diff
@@ -0,0 +1,17 @@
...
+    # Check that variables use square brackets for operations
+    # https://github.com/IAMconsortium/common-definitions/issues/55
+
+    dsd = DataStructureDefinition("definitions/", dimensions=["variable"])
+
+    error = []
+    for variable, attrs in dsd.variable.items():
```
phackstock, line 11: I think you don't need `attrs` here:
```suggestion
    for variable in dsd.variable.keys():
```

phackstock (commented):
Looks good to me in principle, three comments below in line.

definitions/variable/sdg-indicators/sdg.yaml
danielhuppmann: "shape" is a previous project where this variable was already used, so this attribute is used similarly as "navigate" or "engage"...

danielhuppmann: @khaeru suggested in #46 to use a dedicated attribute with sub-attributes for projects, but I suggest to merge this PR as is now and revisit/streamline later...

phackstock: Ah right of course, I read it not as the project. Maybe we should capitalize at some point to make it clearer `SHAPE, ENGAGE, ...`
But good for now.

PR #91, opened 2024-04-11
Add explicit list of components of emissions

danielhuppmann:
This PR adds the list of components for several emissions-variables (where the value is *not* just the sum of sub-components because of the "Energy and Industrial Processes" auxiliary-reporting category.

This list of components is not yet correctly resolved  by **nomenclature** (tags are not replaced and checked against the codelist if they are in a list), but it could still be useful to have this information here as an example for future use.

fyi @IAMconsortium/common-definitions-emissions

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -5,6 +5,12 @@
 - Emissions|{Level-1 Species}:
     description: Emissions of {Level-1 Species}
     unit: "{Level-1 Species}"
+    components:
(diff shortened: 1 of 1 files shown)


phackstock (approved):
Good to be merged from my side. Would probably also be good to open up an issue in nomenclature to support replacing tags in lists.

danielhuppmann:
> Good to be merged from my side. Would probably also be good to open up an issue in nomenclature to support replacing tags in lists.

Done.

PR #92, opened 2024-04-11
Add a general pytest workflow

danielhuppmann:


danielhuppmann:
Thanks @phackstock for noting that the illegal-terms tests wasn't actually running on Github Actions.

Test now failing as expected - and passing again after fixing the intentional violation of the naming conventions.

phackstock (approved):
Thanks for the quick fix. Good to merge.

PR #83, opened 2024-03-22
Add variables for carbon-capture and carbon-removal

jayfuhrman:
per conversation with @danielhuppmann, I have ported most of PR #80 in the tag_carbon-management.yaml file to this new dedicated carbon-management file.  See also PR #82 where I have removed most of what I had previously committed to the carbon-management.yaml file in PR #79 and #80 (and reverted back the tag name) but added several additional energy-consuming CDR tags.  

@strefler @fschreyer @gidden @Renato-Rodrigues

PR #93, opened 2024-04-17
Add a variable capture & removal in the emissions variable tree

danielhuppmann:
Per this issue https://github.com/openENTRANCE/openentrance/issues/333 by @korsbakken, this PR adds a variable to describe (as negative values) the removal of emissions that are not directly linked to an emitter (e.g., DACCS), so that *Emissions|CO2* totals add up.

This PR complements the related discussion on a more fine-grained *Carbon Removal* top-level variable, see #83.

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -10,6 +10,7 @@
       - Emissions|{Level-1 Species}|Energy
       - Emissions|{Level-1 Species}|Industrial Processes
       - Emissions|{Level-1 Species}|Waste
+      - Emissions|{Level-1 Species}|Direct Capture and Removal
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
     definition: Gross emissions of carbon dioxide (CO2), not accounting for negative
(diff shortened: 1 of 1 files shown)


strefler:
Thanks, such a variable makes a lot of sense. I wouldn't highlight DAC in the name or definition though, because it would also include enhanced weathering, ocean CDR, potentially biochar etc. How about "Other Removal"?

danielhuppmann:
I revised the description to make DAC less prominent and also add enhanced weathering for clarity.

About the variable name itself: there is no Emissions variable that has "Removal" in the name, hence "Other Removal" doesn't really make sense. I guess what I meant with the variable name was "Direct Capture and Direct Removal", but repeating "Direct" looks silly...

jayfuhrman (changes requested):
Can we change "Direct Capture and Removal" to simply "Capture and Removal" or "Removal"  or "Atmospheric Removal" ?

danielhuppmann:
Changed as suggested by @jayfuhrman

jayfuhrman:
If this looks good to everyone, we can go ahead and merge?

gidden:
lgtm thanks @danielhuppmann and all involved in discussions!

PR #97, opened 2024-05-17
Add Trade|*|Value variables and fix unit

phackstock:
This PR moves `Trade|*|Value` variables over from legcay-definitions and most importantly fixes the units from `EJ/yr` to `USD_2010/yr`.

FYI @OFR-IIASA

### DIFF
--- a/definitions/variable/macro-economy/trade.yaml
+++ b/definitions/variable/macro-economy/trade.yaml
@@ -8,3 +8,54 @@
     unit: billion USD_2010/yr
     tier: 1
     min: 0
+- Trade|Primary Energy|Biomass|Value:
+    definition: value of net exports of solid, unprocessed biomass, at the global
+      level these should add up to the trade losses only
+    unit: billion USD_2010/yr
+- Trade|Primary Energy|Coal|Value:
+    definition: value of net exports of coal, at the global level these should add
+      up to the trade losses only
+    unit: billion USD_2010/yr
+- Trade|Primary Energy|Gas|Value:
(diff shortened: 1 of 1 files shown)


definitions/variable/macro-economy/trade.yaml
```diff
@@ -8,3 +8,54 @@
...
+- Trade|Secondary Energy|Liquids|Oil [Value]:
+    definition: value of net exports of liquid fuels from petroleum including both
+      conventional and unconventional sources, at the global level these should add
+      up to the trade losses only
+    unit: billion USD_2010/yr
+- Trade|Uranium [Value]:
+    definition: value of net exports of Uranium, at the global level these should
+      add up to the trade losses only
+    unit: billion USD_2010/yr
+- Trade|Emissions Allowances [Value]:
```
orichters, line 59: Is that also "net exports" by model region?

definitions/variable/macro-economy/trade.yaml
```diff
@@ -8,3 +8,54 @@
     unit: billion USD_2010/yr
     tier: 1
     min: 0
+- Trade|Primary Energy|Biomass [Value]:
```
orichters, line 11: Are we sure we want to treat "trade losses" like this? That means the summation checks `World = Sum(Regions)` should be disabled here?

phackstock:
As per discussion with @danielhuppmann, I'll close this PR for now and fix the units directly in legacy-definitions. Once we're ready to move the Trade|*|Value variables we can come back to this.

PR #90, opened 2024-04-10
Add industry variables

macflo8:
This PR is a first draft to add industry and bulk/specialty materials related variables to the data template.

The following variable "categories" are added:

-  Demand of material commodities in total and by demand sector
- Stocks of commodities by sector
- Production of commodities
- Total and collected scrap by commodity
- Trade of material commodities

PR #105, opened 2024-07-25
Add variable for fossil resource extraction

danielhuppmann:
This variable was just requested in another project, and I guess it will be relevant in other projects, too.

### DIFF
--- a/definitions/variable/energy/resource.yaml
+++ b/definitions/variable/energy/resource.yaml
@@ -1,0 +1,3 @@
+- Resource|Extraction|{Primary Fossil Fuel}:
+    definition: Annual extraction of {Primary Fossil Fuel}
+    unit: EJ/yr


masahirosugiyama:
Thank you and I think this proposal ("Resource|Extraction|{Primary Fossil Fuel}") looks great. I am curious, is there any parallel effort to add critical minerals? Anyway, we can easily accommodate other types of variables in a consistent manner and this is good.

danielhuppmann:
@masahirosugiyama, indeed, @macflo8 is working on adding variables for materials-reporting (#90), so critical minerals should be added either as "extraction" (like this PR) or like "production" (similar to that other PR).

Please open a PR so that we can discuss concretely which critical minerals should be added.

danielhuppmann:
Merging, as there were no objections within two weeks.

Issue #39, opened 2023-11-09
Discussion: Add source-level information to final- and secondary energy variables


danielhuppmann:
Most final-energy variables (in ENGAGE, NAVIGATE and other projects) have a split by sector, secondary energy carrier, and primary energy from which that secondary energy carrier was produced, e.g., "Final Energy|Industry|Gases|Biomass“. However, in practice, the source of the secondary-energy carrier cannot be known exactly after the fuel is blended into the transmission system, e.g., electricity in the grid or methane molecules in a pipeline. Including all these combinations leads to an enormous number of variables that only offer spurious precision. Therefore, this sub-categorization was not carried over into this new variable list.

However, @Renato-Rodrigues objected to this exclusion in https://github.com/IAMconsortium/common-definitions/issues/34#issuecomment-1804150484.

This issue is aimed to discuss whether (and which) secondary- and final-energy variables should have information about the source of the energy source.

robertpietzcker:
I think this information (are the liquids used in a specific sector sourced from fossil or biogenic origin) can be important for any study focusing on the EU or sub-levels, where you have different policies targeting eg industry and transport - thus the shares probably will be different across sectors. 

This can also happen with grid-based carriers like gases, hydrogen or heat, as you may have unconnected/partially connected grids. 

So I would make this split part of the variable tree, so that for projects where this information is needed it can easily be supplied and doesn't have to be redefined/discussed again, but to me the variables that detail this on sectoral level would clearly be variables of low priority that I usually would not request teams to report. (only the overall split - so FE|Liquids|Biomass, but not how this splits up into Transport/industry/buildings)

orichters:
I agree that we should rather include these splits in the template and projects can decide not to report them, compared to having different projects reinvent wheels. Maybe we can add a "tier" marking stating which variables are crucial and which might be left out.

Issue #84, opened 2024-03-26
Synthetic Fuel Energy Variables


fschreyer:
Dear all,

we have had internally some discussions about the naming convention of synthetic liquids and gases, which we would like to address on the IAMC-level again. The issue is related to point 2.) raised here #47 .

Currently, we have the following distinctions:
```
- Secondary Energy|Liquids|Biomass:
    description: Production of liquid fuels from biomass
    unit: EJ/yr
- Secondary Energy|Liquids|{Primary Fossil Fuel}:
    description: Production of fossil liquid fuels from {Primary Fossil Fuel}
    unit: EJ/yr
- Secondary Energy|Liquids|Electricity:
    description: Production of liquid fuels from electricity (e-fuels)
    unit: EJ/yr
- Secondary Energy|Liquids|Other:
    description: Production of liquids fuels from sources that do not fit any other category
    unit: EJ/yr
```

``Secondary Energy|Liquids|Electricity`` only includes synthetic fuels where the hydrogen was made via electrolysis from electricity. However, this comes with some problems and ambiguities considering the way synthetic fuels may be implemented in models and how their production may work in the real world. I'd rather opt for something like ``Secondary Energy|Liquids|Hydrogen`` or ``Secondary Energy|Liquids|Hydrogen w/ CCU`` instead. Here are the reasons:

1.) Synthetic fuels are not directly produced from electricity but there is hydrogen as an intermediate product, which is also an energy carrier in the IAMC convention and shows up as ``Secondary Energy|Hydrogen``. It is therefore weird to simply skip this energy carrier, given that the hydrogen needed for synthetic fuels should also be accounted under the SE hydrogen variable (in my view). Also, it is supposedly how some models implement it, i.e. synthetic fuels as a product of hydrogen and captured carbon. At least that's how we do it. 

2.) If models implement it as a conversion from hydrogen and CO2 to liquids, then it is extra effort in the model reporting to split the liquids into those where the hydrogen has been produced from electricity and those where the hydrogen has come from other sources initially (biomass, gas etc.). The latter would then probably need to be reported under ``Secondary Energy|Liquids|Other``. I would find it confusing to interpret model data if outputs of the same technology get attributed to different variables just because the inputs were derived from different sources. I find it much more straightforward to see that a model produces a lot of liquids from hydrogen and then check its hydrogen mix for the sources instead of mixing these two levels. 

There are two counterarguments to this, which come to my mind. 

[...]

PR #106, opened 2024-08-11
Add source-categories for secondary energy carriers

danielhuppmann:
This PR adds final-energy and emissions variables where the source of a secondary energy carrier is explicitly stated.

For clarification - before this PR, only the following variable existed in the common-definitions variable list
> Final Energy|Liquids

This PR will add
> Final Energy|Liquids|Biomass
> Final Energy|Liquids|Oil
> Final Energy|Liquids|Coal
> Final Energy|Liquids|Gas

also for emissions and sectoral disaggregation variables, and also for Solids and Gases.

This PR responds to requests by @robertpietzcker and @orichters in #39 and @OFR-IIASA (offline).

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -102,13 +102,15 @@
     description: Gross emissions of carbon dioxide (CO2) for electricity generation,
       not accounting for negative emissions from bioenergy with CCS (BECCS) in this sector
     unit: Mt CO2/yr
-- Emissions|CO2|Energy|Supply|{Secondary Fuel}:
+- Emissions|CO2|Energy|Supply|{Secondary Fuel with Source}:
     description: Emissions of carbon dioxide (CO2) for production of {Secondary Fuel}
     unit: Mt CO2/yr
-- Gross Emissions|CO2|Energy|Supply|{Secondary Fuel}:
-    description: Gross emissions of carbon dioxide (CO2) for production of {Secondary Fuel},
+    tier: "{Secondary Fuel with Source}"
+- Gross Emissions|CO2|Energy|Supply|{Secondary Fuel with Source}:
+    description: Gross emissions of carbon dioxide (CO2) for production of {Secondary Fuel with Source},
       not accounting for negative emissions from bioenergy with CCS (BECCS) in this sector
     unit: Mt CO2/yr
+    tier: "{Secondary Fuel with Source}"
 - Emissions|CO2|Energy|Supply|Autoproduction:
     description: Emissions of carbon dioxide (CO2) for own-use energy supply
     unit: Mt CO2/yr
--- a/definitions/variable/energy/final-energy.yaml
+++ b/definitions/variable/energy/final-energy.yaml
@@ -13,21 +13,21 @@
 - Final Energy|Non-Energy Use:
     description: Final energy consumption in non-combustion processes
     unit: EJ/yr
-- Final Energy|Non-Energy Use|{Secondary Fuel}:
-    description: Final energy consumption of {Secondary Fuel} in non-combustion processes
+- Final Energy|Non-Energy Use|{Secondary Fuel with Source}:
+    description: Final energy consumption of {Secondary Fuel with Source} in non-combustion processes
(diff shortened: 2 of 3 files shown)


masahirosugiyama:
@danielhuppmann 
Many thanks. This is great. I wonder if we need something like 
Final Energy|Liquids|Synthetic 
We should account for e-fuel. Not sure if Synthetic is the good term.

danielhuppmann:
You're right @masahirosugiyama - curious why no modeling team found it relevant to report this yet.

Following the logic of the other categories (liquids produced from ...), I added this as a category "Liquids|Electricity".

fschreyer:
E-fuels should definitely be included. I would not use the term ``synthetic`` in the variable definition as biofuels or Gas-to-Liquids are also synthetic fuels but rather the source energy carrier as you do. 

However, I would consider alternative definition / naming to ``Liquids|Electricity`` as raised in #84. Or at least it should be made clear that ``SE|Hydrogen`` also includes hydrogen used for ``SE|Liquids|Electricity``.

danielhuppmann:
Thanks @fschreyer for bringing up that important question again,

See a proposal at #108 implemented on the secondary-energy level (to keep the discussion in this PR narrow)- let's discuss there, and once there is agreement, I'll bring over that solution to this PR.

danielhuppmann:
Following the merging of #108, I also added the sub-sub-categories for hydrogen to this PR.

Unless there are any objections, I will merge this PR by the end of this week.

danielhuppmann:
Merging as announced last week.

robertpietzcker:
sorry for coming late to the party (I am only now following up with mails from my vacation time): 
@danielhuppmann could you add one entry that is missing, namely `Gases|Hydrogen`? 
(and potentially the four subcategory of hydrogen)
(and potentially `Gases|Electricity` for symmetry to how Liquids are dealt with)


(It will be fun in the next MIPs to come out with ways to deal with the fact that some models will report `Liquids|Hydrogen` and some models will report `Liquids|Electricity`, but I also don't see a clear way how to improve/merge/change categories)

danielhuppmann:
Yes, good suggestion, thank you. I'll start a PR later today or tomorrow.

pkyle:
I think we should follow the convention that onsite production and use of hydrogen is not reported as such; otherwise we would need to track the intermediate transformations in IGCC power plants and FT synthesis of liquid fuels. In this same way, E-fuels and e-methane should be categorized as taking an input of electricity, unless the model is representing a technology that uses purchased hydrogen. Anyway, I support making all of these categories available in the template, but the user instructions should specify when to use e.g. `Liquids|Electricity` as opposed to `Liquids|Hydrogen`.

danielhuppmann:
@pkyle, in my reading, this guidance is already included in the current description

```yaml
    - Liquids|Hydrogen|Electricity:
        description: liquid fuels using hydrogen from electrolysis, i.e., e-fuels (if
          hydrogen is explicitly modeled as a commodity and not an implicit intermediate
          product of the transformation technologies)
        unit: EJ/yr
    - Liquids|Electricity:
        description: liquid synthetic fuels from electricity, i.e., e-fuels (if hydrogen
          is not explicitly modeled as an intermediate product)
```

See https://github.com/IAMconsortium/common-definitions/blob/cfa3f70d4d7db5e5a02b653f94985d07bfceb6e7/definitions/variable/energy/tag_secondary_energy_carriers_with_sources.yaml#L37

Or do think that this needs to be explained in more detail in the description or somewhere else?

pkyle:
Oh great! I hadn't seen that; my comment was in reference to @robertpietzcker 's note above.

robertpietzcker:
Exactly, the variable definition is clear.

I was just thinking that in my experience, it is a bit arbitrary if a model has efuels implemented as one technology or as two separate technologies. So for the people doing model intercomparison projects / looking at model data across many models (think AR7 database) and wanting to look at efuels, there will always be the need to first add these two variables (which is also not a perfect solution if models use a lot of biomass/fossils to produce hydrogen).

so it was more a comment/raising awareness that the new (clearer) structure sadly doesn't solve all user problems :-)

PR #107, opened 2024-08-11
Add variables for non-energy water withdrawal & consumption

danielhuppmann:
This PR adds variables for non-energy water withdrawal and consumption. Water-related energy variables will be added after merging #106.

FYI @IAMconsortium/common-definitions-coordination @IAMconsortium/common-definitions-land - adding the land team because we do not have a water team...

### DIFF
--- a/definitions/variable/water/water.yaml
+++ b/definitions/variable/water/water.yaml
@@ -1,0 +1,33 @@
+- Water Withdrawal:
+    definition: Total water withdrawals
+    unit: km3/yr
+- Water Withdrawal|Industrial Water:
+    definition: Water withdrawals for the industrial and manufacturing sector
+    unit: km3/yr
+- Water Withdrawal|Municipal Water:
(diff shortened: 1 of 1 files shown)


jonathandoelman:
Hi Daniel,
Good to see that these variables will be added. Two questions:
- I cant find water withdrawal/energy in the main version, is that correct or I am not looking well? 
- What exactly do you need help with? :) 
Jonathan

danielhuppmann:
@jonathandoelman, thanks - I specifically only included the non-energy water variables to keep the PR small, so that a reviewer (like yourself) actually looks carefully (we all tend to just scroll through if lists are very long).

I just invited you to the land team here, please accept the invite to have review-permissions in this repo.

If you think that these variables are ok and are willing to take responsibility for it, then please accept these changes - I will merge and then set up a follow-up PR for the water-energy-related variables.

jonathandoelman (approved):
this looks good to me, nicely in line with previous template versions that I am aware of. thanks!

PR #108, opened 2024-08-12
Add secondary-energy variables for liquids from hydrogen

### LINKED ISSUE #84: Synthetic Fuel Energy Variables

danielhuppmann:
This PR adds categories for production of liquid fuels from hydrogen where hydrogen is modeled explicitly as a (traded) intermediate commodity. This is a proposed compromise for @fschreyer and @pkyle, see #84 

Given the difference on how e-fuels are represented across models, allowing both and letting the modelers (or the project leads where this template is used) make the choice.

closes #84

### DIFF
--- a/definitions/variable/energy/secondary-energy-non-electricity.yaml
+++ b/definitions/variable/energy/secondary-energy-non-electricity.yaml
@@ -17,11 +17,20 @@
     description: Production of liquid fuels using hydrogen from fossils (if hydrogen is
       explicitly modeled as a commodity and not an implicit intermediate product of the
       transformation technologies)
+- Secondary Energy|Liquids|Hydrogen|Biomass:
+    description: Production of liquid fuels using hydrogen generated from biomass
+      (if hydrogen is explicitly modeled as a commodity and not an implicit intermediate
+      product of the transformation technologies)
     unit: EJ/yr
 - Secondary Energy|Liquids|Hydrogen|Electricity:
     description: Production of liquid fuels using hydrogen from electrolysis, i.e.,
       "e-fuels" (if hydrogen is explicitly modeled as a commodity and not an implicit
       intermediate product of the transformation technologies)
+    unit: EJ/yr
+- Secondary Energy|Liquids|Hydrogen|Other:
+    description: Production of liquid fuels using hydrogen from other sources
+      (if hydrogen is explicitly modeled as a commodity and not an implicit intermediate
+      product of the transformation technologies)
     unit: EJ/yr
 - Secondary Energy|Liquids|Electricity:
     description: Production of liquid fuels mainly from electricity, i.e., e-fuels
--- a/definitions/variable/energy/secondary-energy-non-electricity.yaml
+++ b/definitions/variable/energy/secondary-energy-non-electricity.yaml
@@ -17,6 +17,7 @@
     description: Production of liquid fuels using hydrogen from fossils (if hydrogen is
       explicitly modeled as a commodity and not an implicit intermediate product of the
       transformation technologies)
+    unit: EJ/yr
 - Secondary Energy|Liquids|Hydrogen|Biomass:
     description: Production of liquid fuels using hydrogen generated from biomass
       (if hydrogen is explicitly modeled as a commodity and not an implicit intermediate
--- a/definitions/variable/energy/secondary-energy-non-electricity.yaml
+++ b/definitions/variable/energy/secondary-energy-non-electricity.yaml
@@ -12,14 +12,17 @@
     description: Production of liquid fuels from hydrogen (if hydrogen is explicitly
       modeled as a commodity and not an implicit intermediate product of the
       transformation technologies)
+    unit: EJ/yr
 - Secondary Energy|Liquids|Hydrogen|Fossil:
     description: Production of liquid fuels using hydrogen from fossils (if hydrogen is
       explicitly modeled as a commodity and not an implicit intermediate product of the
       transformation technologies)
+    unit: EJ/yr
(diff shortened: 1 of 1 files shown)


fschreyer:
Thanks. I like the compromise.

To have a complete coverage in line with the ``SE|Hydrogen|...`` taxonomy for the case of the intermediate product, I'd propose to also add ``SE|Liquids|Hydrogen|Biomass``, ``SE|Liquids|Hydrogen|Other``. This will allow for complete reporting by the modelers.

fschreyer:
Also, I think we it would be helpful to add to the definition of ``SE|Hydrogen`` -> "Total production of hydrogen including hydrogen used as an intermediate product e.g. for e-fuel production".

masahirosugiyama:
> Also, I think we it would be helpful to add to the definition of `SE|Hydrogen` -> "Total production of hydrogen including hydrogen used as an intermediate product e.g. for e-fuel production". If there are no concerns about that.

I second the suggestion. The difference with FE (finaly energy) would also clarify the amount of an intermediate use (though there would be a loss too).

danielhuppmann:
Thanks for the suggestions, @fschreyer - please start an issue if you think we should add more level of detail

PR #111, opened 2024-08-27
Add variables for food demand

danielhuppmann:
This PR adds the variables for food demand based on the xlsx overview developed by @flohump @realxinzhao and @kanishkan91.

I added references to the agmip variables.

For consistency, I suggest that we also rename this variable by adding "[per capita]" at the end, to highlight that these are per-capita variables.

It's not clear to me whether we need to also add "Food Energy Supply" (which existed in ENGAGE and NAVIGATE), but which seem to be overlapping with sub-categories of Agricultural Production, right?

### DIFF
--- a/definitions/variable/agriculture/food.yaml
+++ b/definitions/variable/agriculture/food.yaml
@@ -2,13 +2,19 @@
     definition: Total food demand in calories
     unit: kcal/cap/day
     weight: Population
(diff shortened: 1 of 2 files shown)


flohump (approved):
Looks good. Food enery demand in EJ can be dropped.

PR #113, opened 2024-08-29
Rework CO2 emissions to only refer to fossil CO2

danielhuppmann:
In NAVIGATE and ENGAGE, the descriptions of the CO2-emissions variables explicitly referred to "fossil CO2" only. This was dropped (my mistake, I guess?) when transferring the variables to the common-definitions repo.

Please advise whether to add "fossil" to all CO2 emissions variable definitions.

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -13,8 +13,8 @@
       - Emissions|{Level-1 Species}|Capture and Removal
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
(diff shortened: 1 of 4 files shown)


jayfuhrman (approved):
This looks good to me.

volker-krey:
Restricting the CO2 emissions to _fossil_ seems wrong to me, in particular as we have subcategories `AFOLU` and `Industrial Processes` which are by definition not _fossil_. Also, negative emission from bioenergy with CCS are not fossil, but are reported under `Emisisons|CO2|Energy` (not under `Gross Emissions` though). So referring to _fossil_ should be avoided here from my perspective.

PR #115, opened 2024-08-30
Add region definitions and mappings for IMAGE  3.4

danielhuppmann:
This PR adds the region definitions for IMAGE 3.4, copy-pasting and making minor adjustments from earlier versions (3.2 in ENGAGE and 3.3 in GEO7 and NAVIGATE) - please double-check.

Two additional requests:
1. In the common-definitions repo, we added all G20 members as possible comparison-regions (see [here](https://github.com/IAMconsortium/common-definitions/blob/main/definitions/region/g20.yaml), please check if any IMAGE regions can be mapped directly to G20 members. I have remove the EU-mapping because this seems to be quite different from actual EU membership (UK, NO) to avoid confusion later on...
2. Please add countries to the region-definitions, see for example from the [AIM 3.0 definitions](https://github.com/IAMconsortium/common-definitions/blob/main/definitions/region/native_regions/AIM_3.0.yaml) - this will allow to do downscaling, etc. automatically later on...


danielhuppmann:
Merging to finalize model registration

PR #116, opened 2024-09-01
Add region definitions and mappings for GCAM 7.1

danielhuppmann:
This PR adds the region definitions and mappings for the GCAM 7.0 model.

I renamed a few of the GCAM-region-names displayed in the Scenario Explorer (and used in scenario ensembles) for better readability and for consistency (in naming if not in the countries with REMIND and IMAGE (e.g. Europe_Non_EU: -> Non-EU Europe.

Still to do, either as comments to this PR or as follow-up:
- Reconsider renaming the European regions for better intuition: EU-12 and EU-15 are not obvious, Non-EU-Europe actually already includes some EU countries and could better be described as South-Eastern (non-EU) Europe.


christophbertram:
Hi Daniel,

Sorry for the slower response. So after also discussing internally, I think our preference would be to for the moment just go with the names we have been using for a number of years still. The mapping from the internal to the display names is ok if you insist, but it actually would make our lives easier still to keep also those display names to the ones we have internally (only with the GCAM 7.1| prefix, please in any case correct from 7.0 to 7.1). This also how it currently is done for GCAM 6.0 in NGFS...
Going forward, we are discussing and will eventually change the EU regions to reflect Brexit and joining of the EU of Croatia. Whether or not this will happen before the final scenarioMIP results or not is also unclear at this point (as is which GCAM version it will be, but a good guess it will be either GCAM 7.2 or 7.x)...

danielhuppmann:
Would a python- or R-script that easily converts scenario data from `GCAM 7.0|Australia and New Zealand` to `Australia_NZ` change your preference?

christophbertram:
Hi Daniel,
No, I do have such a script, and can of course do such a renaming, but of course it adds a potential failure point (as it's easier to overlook one of 32 regions being wrongly mapped, than the risk of doing a pure prefix / getting rid of the prefix wrong...). Again, no strong feelings about this, but I feel if our strategy is to have new GCAM region names soon, then now introducing an intermediate version that just looks different to the outside but is still identical to past use, is not overly helpful in my view...

danielhuppmann:
Bumped to GCAM 7.1 and reverted to the internal GCAM names per your request.

When implementing the new regional resolution and new names in GCAM, please make sure that a version-number clearly reflects a specific regional resolution. There is REMIND 3.1 with 12 regions but also with 21 regions, and this creates a lot of headache and scope for error. So please assign a new version number at the same time as adopting a new regional resolution.

PR #120, opened 2024-09-02
Industry critical minerals

masahirosugiyama:
@danielhuppmann, @macflo8, 

Following the discussion here about the production and extraction in the industry, 
https://github.com/IAMconsortium/common-definitions/pull/105#issuecomment-2268664017
I have created a pull request on production of critical minerals. 

I am not a critical minerals expert, so your feedback is greatly appreciated. I took the list of critical minerals from the IEA reports and websites. 
https://www.iea.org/topics/critical-minerals

I have to admit I am a newbie with GitHub. If I have not followed the GitHub best practice, please let me know.

### DIFF
--- a/definitions/variable/industry/tag_accounting_type.yaml
+++ b/definitions/variable/industry/tag_accounting_type.yaml
@@ -1,0 +1,10 @@
+- Accounting Type:
+    - Volume:
+        description: volume in mass units
+        unit: Mt/yr
+        navigate: "|Volume"
+    - Value:
+        description: value in USD
+        unit: USD_2010/yr
+        navigate: ""
+        
--- a/definitions/variable/industry/production.yaml
+++ b/definitions/variable/industry/production.yaml
@@ -1,0 +1,3 @@
+- Production|Critical Minerals|{Critical Minerals Commodity} [{Accounting Type}]: 
+    description: production of {Critical Minerals Commodity} in {Accounting Type}
+    unit: "{Accounting Type}"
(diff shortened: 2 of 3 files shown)


danielhuppmann:
This looks good to me, but I'd be happy for comments or approval from @IAMconsortium/common-definitions-industry @volker-krey

volker-krey:
Thanks @masahirosugiyama for opening this pull request. Adding critical minerals to the template is very relevant. However, I would not classify the materials as critical minerals as this classification might change over time and also different regions and countries may use different definitions. Using a more stable classification would be useful, for example, Nickel, Copper, Manganese, Lithium are non-ferrous metals and the energy and emissions associated with their production and processing would thus have to be reported under the respective non-ferrous metals sector. Copper and Lithium are actually part of the following [PR](https://github.com/IAMconsortium/common-definitions/pull/154/files#diff-dae2942c267fc21844821513928885d6444d5bf19cf7451669bf108cb40b3c1f). Hope this makes sense.

definitions/variable/industry/tag_accounting_type.yaml
volker-krey, line 1: In [PR](https://github.com/IAMconsortium/common-definitions/pull/154) we refrained from adding the Volume vs. Value accounting for now as the Value accounting would require coordination with the [macro-economy](https://github.com/IAMconsortium/common-definitions/tree/main/definitions/variable/macro-economy) group.

volker-krey:
More broadly, I would suggest to integrate the missing 'critical minerals' dimensions into the structure of #154 after that has been merged.

danielhuppmann:
Closing in favor of #157, which adds the volume-variables for the three critical minerals that weren't added by #154. Per the comment by @volker-krey, better to not add a monetary-value equivalent (for now).

PR #122, opened 2024-09-05
Add model regions and mappings for the MESSAGEix-GLOBIOM 2.1 model

danielhuppmann:
This PR adds the region-definitions and mappings for the MESSAGEix-GLOBIOM 2.1 model (12-region) from NGFS, using the new "countries"-attribute of the nomenclature package.

What's still missing is the R9 and R10 mapping and mappings to (selected) G20 members, see https://github.com/IAMconsortium/common-definitions/blob/cfa3f70d4d7db5e5a02b653f94985d07bfceb6e7/mappings/AIM_3.0.yaml#L95 - please point me to existing projects where that has been used, or add as comment/suggestion directly to this PR.


phackstock:
Per bilateral discussion with @OFR-IIASA, the mapping for R12 -> R10 would combine the following regions:

* China+ (R10) = R12_CHN + R12_RCPA
* Europe (R10) = R12_WEU + R12_EEU

For the R9 mapping things are more complicated, and won't fully work out of the box. For example, R9 defines USA which in MESSAGE R12 is contained in NAM (USA + Canada). We'll discuss further with @volker-krey.

danielhuppmann:
Just tried to implement this for R10, but given that Middle East and Africa aren't separated in the 12-region MESSAGE, we can't map to R10 anyway...

Please review and approve this PR so that I can at least merge the regions and R5 mapping.

phackstock (approved):
Good to be merged from my side.

danielhuppmann:
Added R10 based on a bilateral comment by @OFR-IIASA, also added China as an explicit G20 member.

OFR-IIASA (approved):

PR #123, opened 2024-09-06
Add emissions variable for CEDS harmonization

danielhuppmann:
Based on bilateral conversion with @volker-krey and email exchange with Steve Smith


phackstock:
Merging per discussion with @volker-krey.

PR #125, opened 2024-09-07
added water ecosystems

flohump:
This PR adds water ecosystems.

I don't understand the following categories from the previous template, which I therefore excluded:
Water Ecosystems|Glaciers
Water Ecosystems|Mountains

### DIFF
--- a/definitions/variable/land/water-ecosystems.yaml
+++ b/definitions/variable/land/water-ecosystems.yaml
@@ -3,15 +3,15 @@
     unit: million ha
     tier: 3
 - Water Ecosystems|Lakes:
-    definition: All water ecosystems
+    definition: Lakes
     unit: million ha
     tier: 3
 - Water Ecosystems|Rivers:
-    definition: All water ecosystems
+    definition: Rivers
     unit: million ha
     tier: 3
 - Water Ecosystems|Forests:
-    definition: All water ecosystems
+    definition: Forests
     unit: million ha
     tier: 3
 - Water Ecosystems|Wetlands:
--- a/definitions/variable/land/water-ecosystems.yaml
+++ b/definitions/variable/land/water-ecosystems.yaml
@@ -1,0 +1,24 @@
+- Water Ecosystems:
+    definition: All water ecosystems
+    unit: million ha
+    tier: 3
+- Water Ecosystems|Wetlands:
+    definition: Wetlands
+    unit: million ha
+    tier: 3
+- Water Ecosystems|Wetlands|Peatlands:
+    definition: Peatlands
+    unit: million ha
+    tier: 3
+- Water Ecosystems|Wetlands|Peatlands|Intact:
+    definition: Intact peatlands
+    unit: million ha
+    tier: 3
+- Water Ecosystems|Wetlands|Peatlands|Drained:
+    definition: Drained peatlands
+    unit: million ha
+    tier: 3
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Thanks for this PR. For my own understanding of these variables, are drained peatlands still peatland? Wouldn't they be covered as some "Land Cover" type?

Also, please review the descriptions, there are some copy-paste errors. Also, it could be helpful to separate this into a variable-file and a tag-file to avoid duplicating the unit and tier.

flohump:
> Thanks for this PR. For my own understanding of these variables, are drained peatlands still peatland? Wouldn't they be covered as some "Land Cover" type?

Indeed, drained peatlands are usually used for agriculture (cropland, pasture) or forestry. However, it is important to keep track of drained peatlands because they keep emitting CO2 for decades to centuries. This is different to deforestation, where CO2 emissions occur only once when forest is converted to agricultural land. 
Rewetting of drained peatlands can stop the decay of peat almost immediately. 

> 
> Also, please review the descriptions, there are some copy-paste errors. Also, it could be helpful to separate this into a variable-file and a tag-file to avoid duplicating the unit and tier.

I fixed the copy-paste errors.

danielhuppmann:
> Indeed, drained peatlands are usually used for agriculture (cropland, pasture) or forestry. However, it is important to keep track of drained peatlands because they keep emitting CO2 for decades to centuries. This is different to deforestation, where CO2 emissions occur only once when forest is converted to agricultural land.
Rewetting of drained peatlands can stop the decay of peat almost immediately.

Would it make sense to change to 
> Water Ecosystems|Wetlands|Peatlands|Drained [Memo Item]

and explain that this is often a duplicate of Land Cover variables? I really want to avoid people adding "Land Cover" and "Water Ecosystems" and then being confused that the numbers are not stable over time...

flohump:
> > Indeed, drained peatlands are usually used for agriculture (cropland, pasture) or forestry. However, it is important to keep track of drained peatlands because they keep emitting CO2 for decades to centuries. This is different to deforestation, where CO2 emissions occur only once when forest is converted to agricultural land.
> > Rewetting of drained peatlands can stop the decay of peat almost immediately.
> 
> Would it make sense to change to
> 
> > Water Ecosystems|Wetlands|Peatlands|Drained [Memo Item]
> 
> and explain that this is often a duplicate of Land Cover variables? I really want to avoid people adding "Land Cover" and "Water Ecosystems" and then being confused that the numbers are not stable over time...

It's not a duplicate of Land Cover variables because drained peatland used for agriculture or forestry is in most cases only a small fraction of total cropland, pasture or timber plantation area in a region. This is is because peatland area globally is quite small (~400 Mha) and drained peatlands amount only to about 50 Mha globally, compared to ~1500 Mha cropland and ~3000 Mha pasture area.

Also, the variable "Water Ecosystems|Wetlands|Peatlands|Drained" does not differentiate land-use types like cropland and pasture.

Therefore, I think it is fine as is.

danielhuppmann (approved):
Thanks @flohump!

danielhuppmann:
Sorry, too early with my approval - are you sure that water ecosystems -> forests makes sense? What is that?

flohump:
> Sorry, too early with my approval - are you sure that water ecosystems -> forests makes sense? What is that?

Good catch. I also don't knwo what this could be. I will remove it.

danielhuppmann (approved):
Thanks!

PR #126, opened 2024-09-07
added forestry demand and production

flohump:
Forestry demand and production added to "agriculture" folder.
"agriculture" could be renamed to "AFOLU" for better fit.


danielhuppmann (approved):
Thanks, looks good! Will rename the folder in a follow-up PR.

PR #128, opened 2024-09-09
move natural forest from sdg to land-cover

flohump:
move natural forest from sdg to land-cover to avoid duplicate entries in the full template


danielhuppmann (approved):
Thanks for the clean-up and harmonization!

PR #129, opened 2024-09-09
Added country definitions for IMAGE regions

VassilisDaioglou:
Added country definitions for IMAGE model regions based on: https://models.pbl.nl/image/Region_classification_map


danielhuppmann (approved):
Nice, thank you!

Issue #132, opened 2024-09-12
description tag is not expanded in xlsx export


orichters:
description tag is not expanded in xlsx export

Example:
Final Energy|Non-Energy Use|Chemicals|Gases
EJ/yr
Final energy consumption of {Secondary Fuel} by the {Non-Energy Sector with Source} for non-combustion processes

https://github.com/IAMconsortium/common-definitions/blob/258c943fdbc16f362865367dba68daba50bcac93/definitions/variable/energy/final-energy.yaml#L26-L30

danielhuppmann:
@dc-almeida & @phackstock, can you please check whether this is an issue in nomenclature?

phackstock:
I'll take a look, my hunch is that we currently don't support substitution of multiple different tags.

orichters:
For the variables it works, but not for the description, even if there is only one single tag:
- "Use of {Secondary Fuel} for carbon management by production of biochar using pyrolysis"
- "Use of {Secondary Fuel} for carbon management by direct air capture"
- "Use of {Secondary Fuel} for carbon management by electrochemical separation of dissolved inorganic carbon from seawater"

etc.

orichters:
I think I got the problem. It is the use of a tag in the description that is not present in the variable name as well, see here:
https://github.com/IAMconsortium/common-definitions/blob/258c943fdbc16f362865367dba68daba50bcac93/definitions/variable/energy/final-energy.yaml#L120-L125

phackstock:
Thanks for checking @orichters, I think we have to indeed open a nomenclature issue for that one.

orichters:
In the sense that these errors are identified in the tests?

phackstock:
In the sense that I think there's a bug here. A fix would of course also come with a test.

phackstock:
Ah never mind, the tag replacement logic of nomenclature is sound after all. 
What we could check for is that all tags used in a variable need to occur in the name of the variable.

phackstock:
Closing this issue as the non-replacement of the tags is expected and intended. 
If we would replace tags within a definitions that are not part of the name, we'd get duplicates.

PR #133, opened 2024-09-13
fix occurence of tag in description different from use in variable

orichters:
- fixes https://github.com/IAMconsortium/common-definitions/issues/132

### DIFF
--- a/definitions/variable/energy/final-energy.yaml
+++ b/definitions/variable/energy/final-energy.yaml
@@ -24,7 +24,7 @@
     description: Final energy consumption by the {Non-Energy Sector} for non-combustion processes
     unit: EJ/yr
 - Final Energy|Non-Energy Use|{Non-Energy Sector}|{Secondary Fuel with Source}:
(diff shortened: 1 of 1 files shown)


phackstock (commented):
@orichters, thanks for the fix, was the occurrence of `{Non-Energy Sector with Source}` a mistake in the first place?
If so, good to merge from my side.

orichters:
I think this was an error by Daniel where he added the "with Source" part at the wrong position: https://github.com/IAMconsortium/common-definitions/commit/4f77be867cfa400c41ad826a64ad32e5b13d18ca#diff-fcd91785e0d03088310cced930c54d086a62bf30ae1c102275607adf963ff65aL26

phackstock:
Ah ok, thanks for the clarification. If it's not urgent, I'd rather wait for Daniel before we merge.

orichters:
@danielhuppmann: Can we merge this and fix this error?

danielhuppmann (approved):
Thanks for the fix, @orichters

PR #135, opened 2024-09-13
Adding REMIND-MAgPIE 3.4-4.8 according to current model version

laurinks:
We would like to register the latest REMIND-MAgPIE version. After merging the PR, I would also need the corresponding permission to upload.

Thank you!

### DIFF
--- a/definitions/region/native_regions/REMIND-MAgPIE_3.4-4.8.yml
+++ b/definitions/region/native_regions/REMIND-MAgPIE_3.4-4.8.yml
@@ -1,0 +1,79 @@
+- REMIND-MAgPIE 3.4-4.8:
+    - REMIND-MAgPIE 3.4-4.8|Canada, Australia, New Zealand:
+        countries: [
+          Australia, Canada, Heard Island and McDonald Islands, New Zealand,
+          Saint Pierre and Miquelon
+        ]
+    - REMIND-MAgPIE 3.4-4.8|China and Taiwan:
+        countries: [ China, Hong Kong, Macao, Taiwan ]
+    - REMIND-MAgPIE 3.4-4.8|India:
+        countries: [ India ]
+    - REMIND-MAgPIE 3.4-4.8|Japan:
+        countries: [ Japan ]
+    - REMIND-MAgPIE 3.4-4.8|Other Asia:
+        countries: [
+          Afghanistan, American Samoa, Bangladesh, Bhutan,
+          British Indian Ocean Territory, Brunei Darussalam, Cambodia, Christmas Island,
+          Cocos (Keeling) Islands, Cook Islands, Fiji, French Polynesia,
+          French Southern Territories, Guam, Indonesia, Kiribati, Laos, Malaysia,
+          Maldives, Marshall Islands, Micronesia, Mongolia, Myanmar, Nauru, Nepal,
+          New Caledonia, Niue, Norfolk Island, North Korea, Northern Mariana Islands,
+          Pakistan, Palau, Papua New Guinea, Philippines, Pitcairn, Samoa, Singapore,
+          Solomon Islands, South Korea, Sri Lanka, Thailand, Timor-Leste, Tokelau,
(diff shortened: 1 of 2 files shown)


orichters (approved):
Looks good from my side

danielhuppmann:
Looks ok in principle, but would it be ok to follow the convention from https://github.com/IAMconsortium/common-definitions/blob/main/mappings/REMIND_3.1.yml where the region names only use the REMIND-version number (ie., "REMIND 3.4|...", and the model "REMIND-MAgPIE 3.4-4.8" maps to those regions? This way, we can avoid having multiple definitions/mappings files that are identical.

[Note that the full model version would still always be displayed in the model column in the IAMC format].

orichters:
I would prefer to split up REMIND-MAgPIE and REMIND standalone, as these are really different models, although the default regional split is the same. I see no reason to drop MAgPIE in the region names when the model is used, and the MAgPIE colleagues may even have a stronger opinion on that than I do 🙂

Instead, I wonder whether we can drop all the "3.4" and "3.1" from the REMIND and REMIND-MAgPIE region names, as our development direction is rather to split up existing regions and give them new names, not redefining old ones.

flohump:
> I would prefer to split up REMIND-MAgPIE and REMIND standalone, as these are really different models, although the default regional split is the same. I see no reason to drop MAgPIE in the region names when the model is used, and the MAgPIE colleagues may even have a stronger opinion on that than I do 🙂
> 
> Instead, I wonder whether we can drop all the "3.4" and "3.1" from the REMIND and REMIND-MAgPIE region names, as our development direction is rather to split up existing regions and give them new names, not redefining old ones.

I agree with @orichters. If the coupled REMIND-MAgPIE system is used, the full name "REMIND-MAgPIE" should be used in all instances.

laurinks:
@orichters, @flohump : This would be the version with separate REMIND and REMIND-MAgPIE registration and without version names in the model regions. However, the names of the current versions are listed as model names to enable submission with the specific model version. Is this what you had in mind?

danielhuppmann:
Ok to keep the full REMIND-MAgPIE model name in the region prefix, but please also keep the model version numbers in the prefix.

Also, please use one specific regional Reduktion for each model version - we had plenty of headache when REMIND 2.1 was used with both the 12- and 21-region-versions in one project.

orichters:
REMIND-MAgPIE currently only supports a 12 region version. Therefore, I suggest undoing everything done to the REMIND 3.1 standalone file and just add REMIND-MAgPIE 3.4-4.8 for the 12 region version.

Please discuss possible issues about REMIND 21-region version and its splitting up with @Renato-Rodrigues or other people submitting REMIND 21-region results.

danielhuppmann:
> REMIND-MAgPIE currently only supports a 12 region version. Therefore, I suggest undoing everything done to the REMIND 3.1 standalone file and just add REMIND-MAgPIE 3.4-4.8 for the 12 region version.

Sound good - and I suggest to bump the version number to REMIND-MAgPIE 3.5-4.9 (or higher) when implementing higher regional resolution.

danielhuppmann:
Permissions granted for REMIND modelers to submit scenarios for “REMIND-MAgPIE 3.4-4.8” to the SSP-submission instance.

laurinks:
Thank you, @danielhuppmann ! Permission has been granted. Currently the [submission template](https://files.ece.iiasa.ac.at/ssp-submission/ssp-submission-template.xlsx) still uses REMIND 3.1 leading to the rejection of the updated model name when processing the upload file in the scenario explorer. No problem if it is more efficient for you to wait for other updates of the Excel file. Please let me know once it is updated and I will process the submission.

danielhuppmann:
@phackstock, can you please take a look why the SSP-submission template wasn’t updated upon merging this PR?

phackstock:
@danielhuppmann, it wasn't updated because I pinned it as a precautionary measure. I can remove the pin but then we're playing with fire when it comes to potential variable or unit changes in the future.

PR #137, opened 2024-09-23
Add R10 to GCAM mapping

christophbertram:


danielhuppmann (commented):
Looks good, but do you want to add R9-regions as well for completeness? Also, if you could insert the R10 regions below the R5 mappings, so that all mapping files have the order World - R5 - R9 - R10 - G20…

christophbertram:
Good point, I added R9, and re-ordered...

danielhuppmann (approved):
Thanks!

PR #147, opened 2024-09-26
Add variables for capital stock, investment and trade

danielhuppmann:
Based on an email by @gunnar-pik, I added a further set of macro variables.

### DIFF
--- a/definitions/variable/macro-economy/trade.yaml
+++ b/definitions/variable/macro-economy/trade.yaml
@@ -8,3 +8,52 @@
     unit: billion USD_2010/yr
     tier: 1
     min: 0
+
+# emissions allowances
+- Trade|Emissions Allowances [Volume]:
+    description: Net trade in GHG emission allowances measured in physical quantities
+    unit: Mt CO2-equiv/yr
+    engage: Trade|Emissions Allowances|Volume
+    navigate: Trade|Emissions Allowances|Volume
+- Trade|Emissions Allowances [Value]:
+    description: Net trade in GHG emission allowances measured in monetary quantities
+    unit: billion USD_2010/yr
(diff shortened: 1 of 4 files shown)


gunnar-pik:
Thanks, @danielhuppmann! 
For the investments in electricity / hydrogen / liquids production capacities ..., we will still need to split by energy carrier.

danielhuppmann:
Of course, the more disaggregated variables will be added in a follow-up PR. My experience is that if we add a hundred variables or more in one go, nobody really takes a look - but if we keep the set of changes such that they fit on a page (or so), someone may take a look and notice copy-paste errors or inconsistencies.

danielhuppmann:
Will merge tomorrow morning unless there are any objections

definitions/variable/macro-economy/trade.yaml
```diff
@@ -8,3 +8,52 @@
     unit: billion USD_2010/yr
     tier: 1
     min: 0
+
+# emissions allowances
+- Trade|Emissions Allowances [Volume]:
+    description: Net trade in GHG emission allowances measured in physical quantities
```
orichters, line 14: While I would prefer "Net Exports" instead of "Trade" also here, I think at least the description should indicate this. Same below
```suggestion
    description: Net exports in GHG emission allowances measured in physical quantities
```

orichters (approved):
Small suggestion, looks fine from my side otherwise.

definitions/variable/macro-economy/trade.yaml
danielhuppmann: I don't think exports of allowances makes sense so I instead added a slightly longer description to add clarity.

PR #57, opened 2024-02-15
Add a first list of Tier 1 variables

FlorianLeblancDr:
I open a new pull-request with only Tier 1 variables.
I found it easier for @IAMconsortium/common-definitions-macro-economy to follow discussions and agree with PR with manageable size (working with small subset of variables and from the same tier).

Open to review for @IAMconsortium/common-definitions-macro-economy 

@orichters thank for your work with #56 being usefull for the Tier 2 & 3 as well.

PR #148, opened 2024-09-26
Refactor the goods-trade variable

danielhuppmann:
Based on an email by @gunnar-pik, I realized that the goods-trade variable introduced in #57 deviated from the NAVIGATE and ENGAGE templates by separating imports and exports. I'm not sure if that distinction is strictly necessary?

Also, I suggest to add the Goods category to clarify that this should not include the energy or emissions-allowance trade (see #147).

### DIFF
--- a/definitions/variable/macro-economy/trade.yaml
+++ b/definitions/variable/macro-economy/trade.yaml
@@ -1,10 +1,6 @@
-- Trade|Export [Value]:
-    description: total exports measured in monetary quantities
+- Trade|Goods [Value]:
+    description: Net exports of all goods measured in monetary quantities
     unit: billion USD_2010/yr
     tier: 1
-    min: 0
-- Trade|Import [Value]:
-    description: total imports measured in monetary quantities
-    unit: billion USD_2010/yr
-    tier: 1
-    min: 0
+    navigate: Trade
+    engage: Trade


orichters:
I do not know whether renaming would be an option, but every single time I look at trade variables, I have to check the definitions whether they represent net imports or exports, so I wonder whether naming it `Net Exports|Goods` or `Net Exports|Oil` would not be preferable.

danielhuppmann:
Interesting suggestion - the NAVIGATE template had separate **Export|...** and **Import|...** variables for agriculture, energy, industry and **Trade|...** for emissions-allowances and energy carriers. So instead of switching everything to Trade, we could harmonize to separate Export and Import variables...

danielhuppmann:
@FlorianLeblancDr, given that you added the distinction into Import and Export in #57 (which can't be reported by REMIND), is it ok to revert to just one aggregate variable that follows the logic of previous templates?

danielhuppmann:
Merging to proceed with the ScenarioMIP-SSP update, please start a new PR if you want to discuss a distinction into imports and exports.

FlorianLeblancDr:
Sorry, just seeing this, need to get my github notification sunder my eyes.

Thanks for dealing with this Daniel and Oliver.
Most macroeconomic models (including CGE) have exports and imports separately, I will create a new PR being flexible on reporting 'net' as well.

PR #149, opened 2024-09-26
Add an "International" region

### LINKED ISSUE #74: Role of international bunker emissions / global aggregates
Dear @IAMconsortium/common-definitions-emissions team,

I would be glad if you could decide about the treatment of international bunkers.

As far as I know, In the AR6 template, `Emissions|*` variables for `BC`, `CO`, `CO2`, `NOx`, `OC`, `Sulfur` and `VOC` had the international bunkers added only at the global level, leading to a situation that the sum over all regions did not match `World` results. In the NGFS project, we found out that the MESSAGE team also used this treatment for `Final Energy|Gases` and `Primary Energy|Gas`, while REMIND did it only for Emissions, and GCAM not at all. So there is definitely much room for confusion.

So I would be pleased if the variable description clarified how this should be treated. Thanks!

If I know that correctly, in NAVIGATE, the international bunkers were meant to be distributed to the regions, and I would propose to stick to that but make it explicit in the variable definition.

danielhuppmann:
Per comments by @jkikstra and @byersiiasa, this PR adds an "International" region that can be used for data that cannot be assigned to a specific region, e.g., emissions related to bunker fuels in international aviation and shipping.

Using this region would allow that aggregation of "World = sum of regions + International" works as expected.

closes #74

### DIFF
--- a/definitions/region/common.yaml
+++ b/definitions/region/common.yaml
@@ -1,5 +1,8 @@
 - common:
     - World
+    - International:
+        description: This region should be used if data cannot be assigned to a
+          specific region, e.g., emissions from bunker fuels in aviation or shipping
 
 # The R5 definitions are taken from the call for submissions
 # to the AR6 WG3 scenario ensemble


orichters:
I don't really understand what the purpose of this should be. I understood our discussion in https://github.com/IAMconsortium/common-definitions/issues/74 such that the idea is that international bunkers are allocated to the regions, such that they actually sum to the global value. That is how MESSAGE does it now, as far as I understood @vruijven.

Adding such a region will result in confusion, I fear. You will have a "region" that will only be used for some variables at all. What is supposed to go into it? That would need to be explained very clearly for each single variable. I would rather go with the NAVIGATE / ELEVATE solution to just state that regional values are supposed to sum to World for every variable.

danielhuppmann:
FWIW, this PR doesn't suggest that modeling should switch from their current practice. This is rather intended as a forward-looking option, for future projects - if in some future case, we realize that allocating to "real" regions doesn't make sense (for whatever reason), we already have an "International" region defined that can be used right away without having to re-discuss.

danielhuppmann:
Merging for future use cases.

PR #150, opened 2024-09-26
Add variables for gases from hydrogen

danielhuppmann:
This PR implements the variables for gases-production from hydrogen as suggested by @robertpietzcker in https://github.com/IAMconsortium/common-definitions/pull/106#issuecomment-2328867412.

The implementation follows the distinction into hydrocarbons from electricity (directly) vs. explicitly modelled hydrogen, which was introduced in #106.


danielhuppmann:
Will merge tomorrow morning unless there are any objections.

PR #152, opened 2024-09-27
Add WITCH 6.0 and AIM/HUB-Global 2.4 region definitions

dc-almeida:
Migrating from ssp-workflow repository


danielhuppmann (approved):
Please add empty lines at the of the two files, then good to be merged!

PR #154, opened 2024-10-02
Add industry "Production" variables

macflo8:
This PR is based on a variable and tag subset of #90 with some additional updates. It adds an updated version of the `Production|*` category of industry related variables.

**Changelist:**
- the use of the [Accounting Type tag](https://github.com/IAMconsortium/common-definitions/pull/90/files#diff-843fa5e31a0c5b08908b5723e4237c9f5612d92f59ef458e6cca0d86a524bb4a) has been dropped here. Commodity production expressed in monetary units first requires discussion and agreement with @IAMconsortium/common-definitions-macro-economy
- The concept of [commodity origin](https://github.com/IAMconsortium/common-definitions/pull/90/files#diff-83972244c172d21876f45f36e97374d87575a8c07dcdd4f671524c8ea26f5904) used for metal production has been adjusted for [plastics](https://github.com/IAMconsortium/common-definitions/compare/main...macflo8:industry_production?expand=1#diff-d771e8e3e894620d913a60a581b9892d6a21366b6447f1659eb93190741bee59)
- Several tag descriptions have been extended and improved
- Tier classification system used for emissions has been introduced for commodity production


definitions/variable/industry/tag_plastic_origin.yaml
```diff
@@ -0,0 +1,6 @@
+# Adjusted from metal origin definitions according to ISO IWA 19:2017(E)
```
danielhuppmann, line 1: This tag seems not to be used in the variable definitions.

definitions/variable/industry/tag_iron_commodities.yaml
```diff
@@ -0,0 +1,15 @@
+- Iron Commodity:
```
danielhuppmann, line 1: Is it planned that there will be additional iron commodities? If no, better to avoid this tag-list.

definitions/variable/industry/tag_iron_commodities.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - Steel:
+        description: raw steel, an alloy of carbon and iron
+
+- Iron Intermediate:
```
danielhuppmann, line 5: There can only be one tag-list per yaml file.

definitions/variable/industry/tag_chemical_commodities.yaml
```diff
@@ -0,0 +1,23 @@
...
+          description: the primary chemical with molecular formula of C3H6, second most important precursor for synthetic plastics
+    - High-Value Chemicals|BTX:
+          description: the aromatic compounds Benzene, Toluene, Xylene (BTX)
+    - Ammonia:
+        description: inorganic chemical with the formula NH3, precursor for nitrogen fertilisers
+    - Methanol:
+        description: the simplest alcohol used as precursor for various chemicals and fuels
+
+- Intermediate Chemical:
+    - Formaldehyde:
```
danielhuppmann, line 20: Repeating my earlier comment: what's the difference between Formaldehyde and Formaldehyde Resins, and is there a way of disambiguation?

definitions/variable/industry/production.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+- Production|Chemicals|{Chemicals Commodity}:
+    description: Production of {Chemicals Commodity}
+    unit: Mt
+    tier: 1
+- Production|Chemicals|{Primary Chemical}:
+    description: Production of {Primary Chemical}
+    unit: Mt
+    tier: 1
+- Production|Chemicals|{Chemicals Commodity}|{Metal Origin}:
```
danielhuppmann, line 44: Does the distinction into primary and secondary make sense for chemicals?

definitions/variable/industry/tag_chemical_commodities.yaml
```diff
@@ -0,0 +1,23 @@
+- Chemicals Commodity:
```
danielhuppmann, line 1: Is there a need for separating the chemicals sub-categories into three different tag-lists?

danielhuppmann (commented):
Thanks @macflo8!

See a few comments inline, but overall, I'm confused why you are introducing so many tag-lists. This makes the `productionn.yaml` much longer and complicated to understand that if you simply had one tag-list like
```yaml
- Iron Commodity:
    - Steel:
        description: raw steel, an alloy of carbon and iron
    - Steel|Primary:
        description: raw steel extracted from iron ores and free of reclaimed metal scrap
    - Steel|Secondary:
        description: raw steel that does not directly originate from a primary mineral but from a recycling process or from the processing of waste streams from primary production
    - Iron:
        description: raw iron, used in the iron industry for the production of steel
    - Pig Iron:
        description: pig iron, an intermediate good used in the iron industry for the production of steel produced by smelting iron in a blast furnace
    - Sponge Iron:
        description: sponge iron (also called direct reduced iron)
    - Iron Ore:
        description: ores, from which metallic iron can be economically extracted
```

But 
1. Should pig iron and sponge iron be subcategories of iron?
2. Should iron ore not be categorized as an "Extraction" variable?

macflo8:
Thanks for the feedback, @danielhuppmann.
I understood the tag feature as a way to group variable elements that have some commonalities and to abstract them to some extent. In the case of primary and secondary production, the general definition from ISO applies to all metals. So I thought

```yaml
- Non-Ferrous Metals Commodity:
    - Aluminum:
        description: aluminum, commonly produced by electrolysis of alumina or re-melting of scrap
    - Copper:
        description: copper, a soft and ductile metal with high thermal and electrical conductivity
    - Lithium:
        description: lithium, an alkali metal commonly produced through electrolysis of LiCl, mainly used for electrochemical energy storage
```
```yaml
# Origin of metal according to ISO IWA 19:2017(E)
- Metal Origin:
    - Primary:
        description: extracted from minerals and free of reclaimed metal scrap
    - Secondary:
        description: that does not directly originate from a primary mineral but from a recycling process or from the processing of waste streams from primary production

```
```yaml
- Production|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}|{Metal Origin}:
    description: Production of {Non-Ferrous Metals Commodity} {Metal Origin}
    unit: Mt
    tier: 2
```
makes sense instead of writing out all 6 variables in `production.yaml`.

In the case of iron and steel sector, I agree that it does currently rather complicate things since we have steel as the only commodity with primary and secondary production. I did it only to be consistent with the other sectors.
So maybe this is overengineered and your suggested option to include "Primary" and "Secondary" directly in the commodity tag list is more practical in the end.

Yes, pig iron and sponge iron are subcategories of iron to my understanding. In NAVIGATE and ENGAGE they were not explicitely defined as subcategories. But I guess we could update to `Production|Iron and Steel|Iron|Pig Iron` and `Production|Iron and Steel|Iron|Sponge Iron`?

definitions/variable/industry/tag_chemical_commodities.yaml
macflo8: It is not absolutely needed no. I did it to group commodities according to their "position" in the supply chain and to have a separate group for commodities with primary and secondary production. Alternatively, this can of course also be defined in the same scheme that you suggested for iron commodities above.

definitions/variable/industry/production.yaml
macflo8: It is definitely needed for plastics. Primary chemicals can be produced from secondary feedstocks as well (e.g. chemical recycling of plastic waste), but currently this is not an established process in industry.

definitions/variable/industry/tag_iron_commodities.yaml
macflo8: Sorry, I forgot about this. I assumed this is allowed since nomenclature does apparently parse all tag-lists if I instantiate a `DataStructureDefinition` from this branch.

definitions/variable/industry/tag_iron_commodities.yaml
macflo8: Further refined/manufactured steel commodities could be added to that list in the future I suppose. The worldsteel association reports production of ingots, semis, tubes, plates etc. in their [annual yearbooks](https://worldsteel.org/wp-content/uploads/Steel-Statistical-Yearbook-2018.pdf). But the tag-list can be created once there is more than one product.

definitions/variable/industry/tag_plastic_origin.yaml
macflo8: Ah there is a typo in `production.yaml`. The chemicals should use this tag instead of the metal origin one.

definitions/variable/industry/production.yaml
danielhuppmann: So I'd only use it for plastics now and add it later for other chemicals if this is actually implemented by at least one model

danielhuppmann:
Right, there is no definitive answer on the best complexity of tag lists, but this implementation seems over-engineered to me. And you can also set the tier in the tag-list to distinguish between more and less important variables.

About pig iron and sponge iron: if these two are intended to sum up to `Production|Iron and Steel|Iron`, then I'd introduce the subcategory.

In any case, it would be very useful to add the variable-name from NAVIGATE so that we can implement automated translation and teams have an easier time updating the reporting scripts.

definitions/variable/industry/production.yaml
```diff
@@ -0,0 +1,56 @@
+- Production|Iron and Steel|{Iron Commodity}:
+    description: Production of {Iron Commodity}
+    unit: Mt
+    tier: 1
```
danielhuppmann, line 4: Always useful to include a cross-reference if the variables were renamed compared to earlier projects
```suggestion
    tier: 1
    navigate: Production|Iron and Steel|{Iron Commodity}|Volumne
```

macflo8:
Alright, I have pushed a commit that reverts back to the simpler tag-list scheme. I will keep a backup [branch](https://github.com/macflo8/common-definitions/tree/industry_production_many_tags) with the more complex tag-lists on my fork. 

I have put the `tier` field into the tags as you suggested, but it seems that `nomenclature` throws an error, when a `tier` value is an integer:

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[21], line 1
----> 1 dsd = DataStructureDefinition(path="C:/Users\Florian Maczek\PycharmProjects\common-definitions\definitions", dimensions=["variable"])

File ~\anaconda3\lib\site-packages\nomenclature\definition.py:68, in DataStructureDefinition.__init__(self, path, dimensions)
     65 for dim in self.dimensions:
     66     codelist_cls = SPECIAL_CODELIST.get(dim, CodeList)
     67     self.__setattr__(
---> 68         dim, codelist_cls.from_directory(dim, path / dim, self.config)
     69     )
     71 if empty := [d for d in self.dimensions if not getattr(self, d)]:
     72     raise ValueError(f"Empty codelist: {', '.join(empty)}")

File ~\anaconda3\lib\site-packages\nomenclature\codelist.py:199, in CodeList.from_directory(cls, name, path, config, file_glob_pattern)
    173 @classmethod
    174 def from_directory(
    175     cls,
   (...)
    179     file_glob_pattern: str = "**/*",
    180 ):
    181     """Initialize a CodeList from a directory with codelist files
    182 
    183     Parameters
   (...)
    197 
    198     """
--> 199     code_list = cls._parse_codelist_dir(path, file_glob_pattern)
    200     config = config or NomenclatureConfig()
    201     for repo in getattr(
    202         config.definitions, name.lower(), CodeListConfig()
    203     ).repositories:

File ~\anaconda3\lib\site-packages\nomenclature\codelist.py:288, in CodeList._parse_codelist_dir(cls, path, file_glob_pattern, repository)
    285             code.repository = repository
    286         code_list.append(code)
--> 288 code_list = cls._parse_and_replace_tags(code_list, path, file_glob_pattern)
    289 return code_list

File ~\anaconda3\lib\site-packages\nomenclature\codelist.py:169, in CodeList._parse_and_replace_tags(cls, code_list, path, file_glob_pattern)
    167 # replace tags by the items of the tag-dictionary
    168 for tag_name, tags in tag_dict.items():
--> 169     codes_with_tags = cls.replace_tags(codes_with_tags, tag_name, tags)
    171 return codes_without_tags + codes_with_tags

File ~\anaconda3\lib\site-packages\nomenclature\codelist.py:117, in CodeList.replace_tags(cls, code_list, tag_name, tags)
    115 for code in code_list:
    116     if "{" + tag_name + "}" in code.name:
--> 117         _code_list.extend((code.replace_tag(tag_name, tag) for tag in tags))
    118     else:
    119         _code_list.append(code)

File ~\anaconda3\lib\site-packages\nomenclature\codelist.py:117, in <genexpr>(.0)
    115 for code in code_list:
    116     if "{" + tag_name + "}" in code.name:
--> 117         _code_list.extend((code.replace_tag(tag_name, tag) for tag in tags))
    118     else:
    119         _code_list.append(code)

File ~\anaconda3\lib\site-packages\nomenclature\code.py:121, in Code.replace_tag(self, tag, target)
    118 if isinstance(value, str) and "{" + tag + "}" in value:
    119     # if the the target has the corresponding attribute replace
    120     if attr in target.flattened_dict:
--> 121         mapping[attr] = value.replace(
    122             "{" + tag + "}", getattr(target, attr)
    123         )
    124     # otherwise insert the name
    125     else:
    126         mapping[attr] = value.replace("{" + tag + "}", target.name)

TypeError: replace() argument 2 must be str, not int
```
As far as I can see there are no other tag-lists with tier fields so far, so maybe this has not been considered in `nomenclature` development yet. As a quick hack, I have put the tier values in quotation marks for now. e.g:
```yaml
- Non-Metallic Minerals Commodity:
    - Cement:
        description: a binder substance used for construction
        tier: "1"
```

macflo8:
I will add `navigate`/`engage` fields with the next commit. For some industry sectors, NAVIGATE and ENGAGE templates only contain `Production|{Sector Name}|Volume`, which I find too general since it cannot be inferred, which commodities of the sector are actually included in the aggregate. Therefore I think we should not include them in this PR here. That means there will not be a navigate/engage mapping for several variables.

macflo8:
I have added the engage/navigate fields now. All equivalent NAVIGATE and ENGAGE variables should be mapped now. However since this PR adds some new commodities, the tag feature also creates navigate and engage values for all tags, while some of them are actually not existent in the legacy templates. 
e.g. `Production|Non-Metallic Minerals|Glass` gets the navigate variable `Production|Non-Metallic Minerals|Glass|Volume`, which has not been in the NAVIGATE template because `Glass` is a newly added commodity.

Maybe this could be fixed by implementing a feature in `nomenclature`, that allows to set the navigate/engage value to a keyword that indicates that this tag does not have legacy equivalents and tells `nomenclature` not to create a navigate/engage entry for the variable. But I am not familiar enough with `nomenclature` to judge whether that would be a sensible feature.

definitions/variable/industry/tag_paper_commodities.yaml
```diff
@@ -0,0 +1,7 @@
+- Pulp and Paper Commodity:
+    - Pulp:
+        description: pulp produced from pulpwood, mainly used for paper, board, textiles and cellophane
+        tier: "2"
+    - Paper:
+        description: think sheet material made from cellulose fibers
```
danielhuppmann, line 6: Are you sure that "think sheet" is a thing?

definitions/variable/industry/tag_iron_commodities.yaml
```diff
@@ -0,0 +1,28 @@
...
+        description: pig iron, an intermediate good used in the iron industry for the production of steel produced by smelting iron in a blast furnace
+        tier: "3"
+        navigate: Pig Iron
+    - Iron|Sponge Iron:
+        description: sponge iron (also called direct reduced iron)
+        tier: "3"
+        navigate: Sponge Iron
+    - Iron Ore:
+        description: ores, from which metallic iron can be economically extracted
+        tier: "2"
```
danielhuppmann, line 28: Add empty line at end of file
```suggestion
        tier: "2"

```

definitions/variable/industry/tag_nonfer_commodities.yaml
```diff
@@ -0,0 +1,36 @@
+- Non-Ferrous Metals Commodity:
+    - Aluminum:
+        description: aluminum, commonly produced by electrolysis of alumina or re-melting of scrap, extracted from minerals and free of reclaimed metal scrap
+        tier: "2"
+    - Copper:
```
danielhuppmann, line 5: I suggest to run-order to Aluminum, Aluminum|Primary, Aluminum|Secondary, ...

danielhuppmann (commented):
Almost ready to be merged, thank you.

A few minor suggestions inline, and one request - please add a line break at 88 characters for consistency of the yaml files

definitions/variable/industry/tag_paper_commodities.yaml
macflo8: Ah this is a typo 😅 it should be "thin sheet" (borrowed from the first sentence on [wikipedia](https://en.wikipedia.org/wiki/Paper))

definitions/variable/industry/production.yaml
volker-krey: There is the chemical recycling route to plastics which essentially breaks the plastics up back into base chemicals which then avoids quality losses down the line (compared to mechanical recycling). So in principle having secondary label also combined with chemicals would make sense.

definitions/variable/industry/production.yaml
```diff
@@ -0,0 +1,32 @@
...
+    tier: "{Iron Commodity}"
+    navigate: Production|Iron and Steel|{Iron Commodity}|Volume
+    engage: Production|Iron and Steel|{Iron Commodity}|Volume
+
+- Production|Non-Metallic Minerals|{Non-Metallic Minerals Commodity}:
+    description: Production of {Non-Metallic Minerals Commodity}
+    unit: Mt
+    tier: "{Non-Metallic Minerals Commodity}"
+    navigate: Production|Non-Metallic Minerals|{Non-Metallic Minerals Commodity}|Volume
+    engage: Production|Iron and Steel|{Iron Commodity}|Volume
```
danielhuppmann, line 13:
```suggestion
    engage: Production|Non-Metallic Minerals|{Non-Metallic Minerals Commodity}|Volume
```

definitions/variable/industry/production.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+- Production|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}:
+    description: Production of {Non-Ferrous Metals Commodity}
+    unit: Mt
+    tier: "{Non-Ferrous Metals Commodity}"
+    navigate: Production|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}|Volume
+    engage: Production|Iron and Steel|{Iron Commodity}|Volume
```
danielhuppmann, line 20:
```suggestion
    engage: Production|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}|Volume
```

definitions/variable/industry/production.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+- Production|Chemicals|{Chemicals Commodity}:
+    description: Production of {Chemicals Commodity}
+    unit: Mt
+    tier: "{Chemicals Commodity}"
+    navigate: Production|Chemicals|{Chemicals Commodity}|Volume
+    engage: Production|Iron and Steel|{Iron Commodity}|Volume
```
danielhuppmann, line 27:
```suggestion
    engage: Production|Chemicals|{Chemicals Commodity}|Volume
```

definitions/variable/industry/tag_chemical_commodities.yaml
```diff
@@ -0,0 +1,44 @@
...
+          ethylene, propylene, butylenes and aromatics
+        tier: "1"
+    - High-Value Chemicals|Ethylene:
+        description: ethylene, most important precursor for synthetic plastics
+        tier: "2"
+    - High-Value Chemicals|Propylene:
+        description: propylene, second most important precursor for synthetic plastics
+        tier: "2"
+    - High-Value Chemicals|BTX:
+        description: the aromatic compounds Benzene, Toluene, Xylene (BTX)
```
danielhuppmann, line 20:
```suggestion
        description: the aromatic compounds Benzene, Toluene and the three xylene isomers (BTX)
```

danielhuppmann (approved):
There are some small issues, then good to be merged.

danielhuppmann:
@macflo8, I took the liberty to commit my review-suggestions directly to your branch so that I can go ahead and merge your PR.

PR #162, opened 2024-10-07
Revert change to fossil-CO2 in emissions variables

danielhuppmann:
As discussed by email and stated by @volker-krey in https://github.com/IAMconsortium/common-definitions/pull/113#issuecomment-2396250910, this PR reverts the restriction of CO2-emissions to only refer to fossil sources (#113).

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -13,8 +13,8 @@
       - Emissions|{Level-1 Species}|Capture and Removal
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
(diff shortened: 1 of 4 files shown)


jkikstra:
Maybe @jayfuhrman wants to weigh in, since I believe he suggested the change in the first place?

danielhuppmann:
> Maybe @jayfuhrman wants to weigh in, since I believe he suggested the change in the first place?

No, I have to admit that this was my doing because I referred back to the NAVIGATE and ENGAGE definitions.

Just need an approval from someone competent to merge this PR.

strefler (approved):

PR #170, opened 2024-10-17
Add general investment variables

danielhuppmann:
This PR adds variables suggested by @gunnar-pik related to general investment. including adding tiers to already-existing variables.

### DIFF
--- a/definitions/variable/macro-economy/investment.yaml
+++ b/definitions/variable/macro-economy/investment.yaml
@@ -2,9 +2,17 @@
     description: Total economy-wide investments including macro-economic capital stock,
       energy system, R&D, etc.
     unit: billion USD_2010/yr
+    tier: 1
+
 - Investment|Agriculture:
     description: Investment related to agricultural activity
     unit: billion USD_2010/yr
+    tier: 1
+- Investment|Forestry:
+    description: Investment into growing forests
+    unit: billion USD_2010/yr
+    tier: 2
+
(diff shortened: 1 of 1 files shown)


FlorianLeblancDr:
Second comment : my problem with the infrastructure variable is that energy supply is part of the infrastruture in some definitions (https://outlook.gihub.org/) 

So may be :
- rename as "Non-energy infrastructure" and be explicit about the list of infrastrucutre we are talking about ? (Water, Transport, Télécommunications)

gunnar-pik:
Having second thoughts about `Investment|Energy Efficiency`: These are clearly part of Energy Demand, and should therefore be part of the substructure of `Investment|Energy Demand|...`. For Buildings, for instance, they might be part of retrofits. So I think we can simply leave the high-level `Investment|Energy Efficiency` variable away

danielhuppmann:
@gunnar-pik, I'm confused about your comment regarding energy-efficiency investment - this is not part of this PR.

Should I go ahead and merge the two variables in this PR: forestry-investment and infrastructure-investment, or do we need a more in-depth discussion about energy-supply-infrastructure versus other infrastructure?

gunnar-pik:
You're right, please go ahead with the merge. 
We can revisit the issue on energy efficiency investments when we add additional investments for the demand side.

FlorianLeblancDr (changes requested):

PR #172, opened 2024-10-18
Refactor list of carbon-removal variables

danielhuppmann:
This PR implements the list of carbon-removal variables as sent by @strefler by email - however, comparing to the already-existing list of carbon removal variables, the following options are not included in the updated template.

Please advise whether these should indeed be removed or whether they should be added to one of the lists?

```yaml
- Peatland and Wetland Restoration:
    description: restoration of peat and coastal wetlands
- Bio-Oil Storage:
    description: use of bio-oil
- Mineral Products:
    description: use of mineral products for carbon sequestration
- Biomass Burial:
    description: burial of biomass on land
- Biomass Sinking:
    description: sinking of biomass in the ocean
- Other Methods:
    description: other methods not explicitly listed
```

fyi @IAMconsortium/common-definitions-coordination


danielhuppmann:
Two changes compared to the list:
1. I renamed "Materials" to "Long-Lived Materials" for consistency with the Carbon-Utilization variables.
2. I renamed "Carbon-Neutral Fuels" to "Synthetic Fuels" in line with the description and to increase clarity

flohump:
> This PR implements the list of carbon-removal variables as sent by @strefler by email - however, comparing to the already-existing list of carbon removal variables, the following options are not included in the updated template.
> 
> Please advise whether these should indeed be removed or whether they should be added to one of the lists?
> 
> ```yaml
> - Peatland and Wetland Restoration:
>     description: restoration of peat and coastal wetlands
> - Bio-Oil Storage:
>     description: use of bio-oil
> - Mineral Products:
>     description: use of mineral products for carbon sequestration
> - Biomass Burial:
>     description: burial of biomass on land
> - Biomass Sinking:
>     description: sinking of biomass in the ocean
> - Other Methods:
>     description: other methods not explicitly listed
> ```
> 
> fyi @IAMconsortium/common-definitions-coordination

I agree on the suggested list of CDR options in this PR.
In particular, I also agree on removing the options listed above. 
I have two requests:
- Please relabel "Afforestation" to Re-Afforestation
- If possible, please add two sub-categories for "Soil Carbon Management": "Soil Carbon Management|Cropland" and "Soil Carbon Management|Grassland"

danielhuppmann:
Thanks @flohump - I added the sub-categories for soil carbon management.

About changing the label to "Re-Afforestation", can you please explain? Is that intended to describe Re- and Afforestation, or afforestation only on areas that previously had forests?

definitions/variable/emissions/tag_geological-storage.yaml
```diff
@@ -0,0 +1,12 @@
+- Geogological Storage Source:
+    - Bioenergy:
+        description: geological storage of CO2 from biogenic orgin, i.e.,
+          bioenergy with carbon capture and storage (BECCS)
+    - Direct Air Capture:
+        description: direct air capture in combination with carbon-capture and
+          sequestration (CCS)
+    - Synthetic Fuels:
```
gidden, line 8: It would be useful to provide more context here in terms of how this is different than bioenergy (bioenergy-based synthetic fuels are also present in scenarios)

definitions/variable/emissions/tag_geological-storage.yaml
```diff
@@ -0,0 +1,12 @@
...
+        description: geological storage of CO2 captured from burning synthetic fuels
+          created from carbon-capture-and-utilization (CCU)
+    - Other:
+        description: geological storage of CO2 from other sources
```
gidden, line 12: If we are putting this under the tier-1 variable `Carbon Removal` then we need to clarify that this does *not* include storage of carbon from fossil sources

definitions/variable/emissions/tag_land-removal-options.yaml
```diff
@@ -0,0 +1,16 @@
+- Land Removal Option:
+    - Agroforestry:
+        description: agroforestry
```
gidden, line 3: Maybe @flohump can provide a bit more text here to describe agroforestry

definitions/variable/emissions/tag_land-removal-options.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - Afforestation:
+        description: afforestation
+    - Biochar:
+        description: conversion of biomass to recalcitrant carbon via pyrolysis and its
+          subsequent application to croplands to increase soil carbon stocks
+    - Soil Carbon Management:
+        description: soil carbon management techniques
```
gidden, line 10: This would be useful to expand. Even a basic description like "enhancing the carbon content of soils through ..."

definitions/variable/emissions/tag_land-removal-options.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - Soil Carbon Management|Cropland:
+        description: soil carbon management techniques on cropland
```
gidden, line 12: perhaps call back to upper level variable where more context will be provided: "See `Soil Carbon Management` for more information"

definitions/variable/emissions/tag_land-removal-options.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - Soil Carbon Management|Grassland:
+        description: soil carbon management techniques on grassland
```
gidden, line 14: perhaps call back to upper level variable where more context will be provided: "See `Soil Carbon Management` for more information"

definitions/variable/emissions/tag_long-lived-materials.yaml
```diff
@@ -0,0 +1,7 @@
+- Long-Lived Materials:
+    - Timber:
+        description: long-lived wood products
```
gidden, line 3: It would be useful to provide indicative time frames (e.g., "timber in construction and other wood products which lasts more than 100 years without decomposing")

definitions/variable/emissions/tag_long-lived-materials.yaml
```diff
@@ -0,0 +1,7 @@
...
+    - Plastics:
+        description: long-lived plastics
```
gidden, line 5: It would be useful to provide indicative time frames (e.g., "bio-based plastics in manufacturing and other material uses which lasts more than 100 years without decomposing")

definitions/variable/emissions/tag_long-lived-materials.yaml
```diff
@@ -0,0 +1,7 @@
...
+    - Other:
+        description: other long-lived materials
```
gidden, line 7: again, timeframes eg "other long-lived materials storing carbon for at least 100 years"

definitions/variable/emissions/tag_ocean-removal-option.yaml
```diff
@@ -0,0 +1,7 @@
+- Ocean Removal Option:
+    - Ocean Alkalinity Enhancement:
+        description: increasing ocean alkalinity
```
gidden, line 3: "increasing ocean uptake of carbon through anthropogenic alkalinity addition"

definitions/variable/emissions/tag_ocean-removal-option.yaml
```diff
@@ -0,0 +1,7 @@
...
+    - Direct Ocean Capture:
+        description: direct ocean capture
```
gidden, line 5: "direct removal of CO2 from the ocean through electrodialysis or other methods"

definitions/variable/emissions/tag_ocean-removal-option.yaml
```diff
@@ -0,0 +1,7 @@
...
+    - Fertilization:
+        description: ocean fertilization and artificial upwelling
```
gidden, line 7: "Stimulation of the primary production of phytoplankton and other organisms which contribution to ocean uptake of CO2 through direct nutrient addition or artificially upwelling deep nutrient-rich water"

gidden (changes requested):
Hi all - this looks generally looks like a good approach to me. I have two broad suggestions:

1. we need to be careful to highlight that carbon removal variables do not include fossil-based carbon sources - I've tried to add comments where I saw potential uncertainties here
2. it would be useful to add more documentation to some variables. I tried adding suggestions where I thought I could. It would be great for others to comment, e.g., @jayfuhrman @flohump or @strefler

definitions/variable/emissions/tag_geological-storage.yaml
danielhuppmann: I switched from "Carbon-Neutral" to "Synthetic" to be less ambiguous... Plus there's a category "Bioenergy". Do you think clearly adding "synthetic fuels not including biofuels" would be sufficient? Or do you have a better suggestion for changing the variable name or extending the description?

definitions/variable/emissions/tag_geological-storage.yaml
danielhuppmann: Added a half-sentence.

definitions/variable/emissions/tag_land-removal-options.yaml
danielhuppmann: Added the first part of your sentence.

definitions/variable/emissions/tag_long-lived-materials.yaml
danielhuppmann: Added the same sentence as in https://github.com/IAMconsortium/common-definitions/blob/be138400d429363066c497535a1ab5af734f9042/definitions/variable/emissions/tag_carbon-utilization.yaml#L5

definitions/variable/emissions/tag_long-lived-materials.yaml
danielhuppmann: Added the same sentence as in https://github.com/IAMconsortium/common-definitions/blob/be138400d429363066c497535a1ab5af734f9042/definitions/variable/emissions/tag_carbon-utilization.yaml#L5

definitions/variable/emissions/tag_ocean-removal-option.yaml
danielhuppmann: Done.

definitions/variable/emissions/tag_ocean-removal-option.yaml
danielhuppmann: Done

definitions/variable/emissions/tag_long-lived-materials.yaml
danielhuppmann: Added the same sentence as in https://github.com/IAMconsortium/common-definitions/blob/be138400d429363066c497535a1ab5af734f9042/definitions/variable/emissions/tag_carbon-utilization.yaml#L5

definitions/variable/emissions/tag_geological-storage.yaml
fschreyer: I would propose the following:

```
Bioenergy:
  description: geological storage of CO2 captured from bioenergy, i.e. energy carriers whose energy content is derived from biomass. This is also referred to as bioenergy with carbon capture and storage (BECCS). 
 [...]
 Synthetic Fuels:
   description: geological storage of CO2 captured from burning synthetic fuels produced from carbon-capture-and-utilization (CCU)
```

definitions/variable/emissions/tag_land-removal-options.yaml
flohump: "agroforestry intentionally integrate trees with crops or pasture on the same land"

definitions/variable/emissions/tag_land-removal-options.yaml
```diff
@@ -0,0 +1,16 @@
+- Land Removal Option:
+    - Agroforestry:
+        description: agroforestry
+    - Afforestation:
+        description: afforestation
```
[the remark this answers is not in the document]
flohump: @danielhuppmann already renamed it to `Re/Afforestation`
We will have these two subcategories in land cover:
`Land Cover|Forest|Planted|Plantation|Re/Afforestation`
    Plantations for carbon sequestration, Reforestation and/or Afforestation with monoculture and/or non-native species
`Land Cover|Forest|Planted|Natural|Re/Afforestation`
    Reforestation and/or Afforestation for carbon sequestration with native tree species

We could add similar subcategories here for CDR but I guess that not many teams will be able to report on this level of detail.

danielhuppmann:
Thank you @gidden  and @flohump, I think I have incorporated all suggestions, please give it one more look...

definitions/variable/emissions/tag_land-removal-options.yaml
```diff
@@ -0,0 +1,16 @@
+- Land Removal Option:
+    - Agroforestry:
+        description: intentionally integrating trees with crops or pasture on the same land
+    - Re/Afforestation:
```
gidden, line 4: Still think this needs a subcategory. I tried to commit a PR, but it didn't work through the web interface:

```
    - Re/Afforestation:
        description: regrowing forested area on previously forested area (natural forest) or growing forested area on previously non-forested area (monoculture/non-native species)
    - Re/Afforestation|Reforestation:
        description: regrowing forested area on previously forested area (natural forest)
    - Re/Afforestation|Afforestation:
        description: growing forested area on previously non-forested area (monoculture/non-native species)
```

gidden: Separate question - how does a `\` work? Would we want to use something else (e.g. `-`)...

danielhuppmann: Well, @flohump initially suggested "Re-Afforestation", but I vetoed this because it can be very confusing (like reforesting only on previously afforested areas?) - see https://github.com/IAMconsortium/common-definitions/pull/180#issuecomment-2443213638 and subsequent preference for `/`.

About the additional sub-categories, I suggest to merge this PR as is, wait for the parallel PR on AFOLU variables to be merged, and then implement a parallel solution in he carbon-removal variable tree.

gidden: @flohump do you agree on this sub-categorization? Can we expect a parallel PR as @danielhuppmann suggests?

flohump: I think it would be more meaningful to split the sub-categories based on the type of Re/Afforestation (natural forest vs. plantation), which is more informative in particular for biodiversity. 
We will have the same differentiation in the land cover variables. 
My suggest would be: 
Re/Afforestation:
        description: Reforestation and/or Afforestation for carbon sequestration
Re/Afforestation|Natural Forest:
        description: Reforestation and/or Afforestation for carbon sequestration with native tree species 
      resembling natural vegetation
Re/Afforestation|Plantation:
        description: Reforestation and/or Afforestation for carbon sequestration with non-native 
      species and/or as monoculture plantation

danielhuppmann:
Implemented the split of Re/Afforestation between natural forest and plantation as suggested by @flohump.

@gidden, good to be merged?

strefler:
> Two changes compared to the list:
> 
>     1. I renamed "Materials" to "Long-Lived Materials" for consistency with the Carbon-Utilization variables.
I added subcategories "long-lived" and "short-lived" to be able to account e.g. for plastics that may last only <5 years. Makes sense?

definitions/variable/emissions/carbon-removal.yaml
```diff
@@ -2,8 +2,71 @@
...
+
+- Carbon Removal|Geological Storage:
+    description: Gross removals of carbon dioxide (CO2) from atmospheric origin
+      or biomass through geological storage
+    unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+- Carbon Removal|Geological Storage|{Geogological Storage Source}:
+    description: Gross removals of carbon dioxide (CO2) through {Geogological Storage Source}
```
strefler, line 17: Do we really want to report gross removals here? I would rather suggest net removals to avoid confusion. E.g. when a high-temperature DACCS plant is using gas to heat, I would subtract the resulting remaining fossil emissions from the DACCS value.

definitions/variable/emissions/carbon-removal.yaml
```diff
@@ -2,8 +2,71 @@
...
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+- Carbon Removal|Land Use|{Land Removal Option}:
+    description: Gross removals of carbon dioxide (CO2) through {Land Removal Option}
+    unit: Mt CO2/yr
+    tier: 2
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+
+- Carbon Removal|Long-Lived Materials:
```
strefler, line 37: I'm a bit torn here. What about materials with a lifetime of 5 years, e.g. plastics? They would have to be reported as removals in one timestep, and emissions in the next. So would rather stick with only materials here, even though I see the appeal of adding "long-lived"

definitions/variable/emissions/carbon-removal.yaml
```diff
@@ -2,8 +2,71 @@
...
+    description: Gross removals of carbon dioxide (CO2) through {Long-Lived Materials}
+    unit: Mt CO2/yr
+    tier: 2
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+    navigate: Carbon Removal|Materials|{Long-Lived Materials}
+    engage: Carbon Removal|Materials|{Long-Lived Materials}
+
+- Carbon Removal|Ocean:
+    description: Gross removals of carbon dioxide (CO2) through ocean sinks
```
strefler, line 55: Also here I would report net removals instead of gross. Especially for ocean liming, there is the possibility to do it without CCS, which would result in rather large gross emissions and removals, but relatively smaller net removals. Reporting only gross removals would be misleading. We can add gross removals later to also have this information

definitions/variable/emissions/tag_geological-storage.yaml
strefler: I recently suggested "Carbon-neutral Efuels" to avoid confusion with biofuels (that are sometimes also categorized as synthetic fuels) as well as Efuels from fossil sources. Another option would be to only call them Efuels, and move the carbon-neutral to the description

definitions/variable/emissions/carbon-removal.yaml
gidden: I would have to disagree here (and therefore by proxy also disagree with the point above). If I am an ocean modeller and I want to understand what is the carbon removal due to OAE based on the alkalinity addition, how would I do this if we mix in the process emissions?

I think we need to report them both (removal through alkalinity addition and emissions generated through creating the alkalinity in the first place). 

There is an interesting argument *where* the positive emissions should live. Should they be reported in the "Carbon Removal" tree? Or should they be part of industry emissions (since this would be an industrial process)?

gidden: Another argument: how do we calculate total cumulative stored carbon? If we report only net, we will underestimate what the model is actually putting underground.

danielhuppmann:
You are commenting on a PR that was merged already. Please start a new issue or PR to discuss changes of the current status of the variable template.

PR #177, opened 2024-10-21
Add capacity (addition) variables for electricity generation

danielhuppmann:
This PR adds investment variables suggested by @gunnar-pik for capacity and capacity additions of electricity generation.

One clarifying question: when I read this description, the capacity-addition variable should be reported as net-change (not just new investment). So if 1 GW of wind turbines are built and 0.2 GW of wind turbines are decommissioned in the same year (due to age), the variable should be reported as 0.8 GW.

Is my understanding correct, and should this be made more explicit in the description or notes?

### DIFF
--- a/definitions/variable/energy/secondary-energy-electricity.yaml
+++ b/definitions/variable/energy/secondary-energy-electricity.yaml
@@ -12,13 +12,13 @@
     description: Installed (available) capacity to generate electricity from {Electricity Source}
     unit: GW
 - Capacity Addition|Electricity:
-    description: Total annual additions to installed (available) electricity generation capacity
+    description: Total annual new installation of electricity generation capacity
     unit: GW
-    notes: This variable should be computed as yearly average additions between previous
-      and current reported time step.
+    notes: This variable should be computed as yearly average additions between the previous
+      and the current reported time step.
 - Capacity Addition|Electricity|{Electricity Source}:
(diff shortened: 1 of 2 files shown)


masahirosugiyama:
@danielhuppmann Thanks a lot. Good catch. Can we add "net of decommissioning / depreciation" or something like that in the definition? In emissions, we have gross emissions and net emissions. I am not sure which is more accepted for capacity additions, but this should be clarified.

danielhuppmann:
Thanks - and for what it's worth, I personally don't think that the formulation "net of ..." is always clear. Maybe someone can propose a completely unambiguous formulation.

pkyle:
Thanks for the clarification--I don't think anyone from my research group has ever estimated net investment, i.e. new capacity additions minus retirements. This variable would not be useful to any of the research questions for which we have calculated power sector investment, as there is no financial recovery from the retirement of equipment. In your view, what is the purpose of estimating net investment? Also it seems that this value could just be calculated as the capacity in one period minus the capacity in the prior time period, so it shouldn't need its own reporting variable, right?

gunnar-pik:
I agree with @pkyle - and the point of the <Capacity Additions|...> variable should precisely be to also get gross additions as an indicator of effort. Net additions can easily be calculated via differences of <Capacity...> over time.

PR #178, opened 2024-10-21
revision of AFOLU variables

flohump:
This revision of reporting variables has been discussed and agreed with representative of all modeling teams in @IAMconsortium/common-definitions-land 
The revision is based on https://1drv.ms/x/s!AlN2Seu8OCz8hLQ8Wh_ysgHyDSg4QA
For consistency with naming conventions some variables have been named and defined differently.
For the tier-levels of Emissions I need some help. E.g. "Emissions|CH4|AFOLU" should be Tier-1 but "Emissions|Kyoto Gases|AFOLU" should be Tier-2. How to do this in the existing structure?


danielhuppmann:
Thanks @flohump, but can you please split this up into several PRs? It's impossible to review if there are hundreds of variables changed simultaneously. Also I see that you changed variables where other topic-groups may also want to chime in, so having targeted PR facilitates a more productive discussion...

danielhuppmann:
> I could roughly split it into one PR per subfolder. Does that make sense?

Yes, please.

PR #182, opened 2024-10-22
LU energy

flohump:
This revision of reporting variables is part of a larger set of changes related to AFOLU reporting variables that have been discussed and agreed with representative of all modeling teams in @IAMconsortium/common-definitions-land 
The revision is based on https://1drv.ms/x/s!AlN2Seu8OCz8hLQ8Wh_ysgHyDSg4QA
For consistency with naming conventions some variables have been named and defined differently.

### DIFF
--- a/definitions/variable/energy/primary-energy.yaml
+++ b/definitions/variable/energy/primary-energy.yaml
@@ -7,79 +7,55 @@
       and forestry residue, municipal solid waste bioenergy and traditional biomass
     unit: EJ/yr
     tier: 1
-    domain: Energy
-    subdomain: PE Biomass
 - Primary Energy|Biomass|{CCS}:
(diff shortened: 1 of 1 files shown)


danielhuppmann (changes requested):
Thank you @flohump, but as written by email separately, I am against adding domain and subdomain attributes explicitly without prior discussion and some guidance on the structure.

If the aim is filtering, this can be easily implemented by something like
```
[v for v in variable_list if "Energy" in v and "Biomass" in v]
```

[And if we really add these attributes, I'd be against using abbreviations]

For now, I suggest to remove the new attributes.

flohump:
Attributes removed.

danielhuppmann (approved):
Looks good, thanks!

Issue #187, opened 2024-10-25
Labelling default `climate-assessment` variables


jkikstra:
We could add a label to the metadata for all variables taken as input by the automated `climate-assessment`/`climate-processor` workflow.

Issue #186, opened 2024-10-25
Labelling IPCC sectors in emissions variables


jkikstra:
Opening this issue to keep track of it.

Would be useful to have IPCC sector labelling in the emissions variables. 

I would suggest to have this as a (i.e., one) meta column, with IPCC categories specified as a string.
This is not perfectly machine-readable, but a step forward from only having it in the variable definition string.

h/t @OFR-IIASA 

The 2019 refinement of the reporting that we can base it on is here: https://www.ipcc-nggip.iges.or.jp/public/2019rf/pdf/1_Volume1/19R_V1_Ch08_Reporting_Guidance.pdf
(please do say so if there's any better source to base it on)

jkikstra:
Note that #188 already implements a few, but kept it to 

Maybe we want multiple labels?
- EDGAR
- IPCC1996
- IPCC2006

Issue #185, opened 2024-10-25
How to best add a "CMIP7-harmonization" tag for a specific list of emissions variables


jkikstra:
I would like to basically implement tags for a variable list very similar to what is available [here [sheet: IAMC variables]](https://docs.google.com/spreadsheets/d/1HKfl21NX8wohPEHpHlmf5XithknPw8zjLpVe4MiOU9U/edit?usp=sharing).

I would like to create a meta column "CMIP7-harmonization" with the options:
"global"
"regional"

These should apply to a list which is a combination of species and sectors.
It is possible to update it later.

I'm wondering how to best implement this, without messing up the current "Level-1/2/3 species" structure.

Issue #184, opened 2024-10-24
Question about sectors under `Final Energy|Transportation`, and consistency with `Emission` sectors.


korsbakken:
The sectors under `Final Energy|Transportation` are very different from what was used in the variable definitions for the [global sectoral pathways in AR6 WGIII](https://data.ece.iiasa.ac.at/ar6-scenario-submission/static/files/global_sectoral.zip). Also, there aren't coresponding sectors under `Emissions|{species}|Energy|Demand|Transportation`.

From #36 , it looks like the current variables under `Final Energy|Transportation` were imported from NAVIGATE and ENGAGE? But many models adopted the AR6 variables, and it's a bit of a tough sell to tell them to switch, when it's not quite clear what the advantage of the current sectors in `common-definitions` is. The structure can also be difficult to switch to if you are currently using the AR6 categories. There is no equivalent of, e.g., `Final Energy|Transportation|Road`. Instead, you are forced to disaggregate into `Final Energy|Transportation|Bus`, `Light Duty Vehicle`, and `Truck`, which might not even be represented in all models that need a `Road` variable.

The second issue is that there are no subsetors at all under `Emissions|{species}|Energy|Demand|Transportation`.

Could we consider either adopting a more AR6-like structure under `Final Energy|Transportation`, or at least have some aggregated variable corresponding to road transport (and preferably also aviation and maritime/navigation including both domestic and international, cf. #36 )?

And regardless of which sector structure is used under `Final Energy|Transportation`, could the same structure also be added under `Emissions`?

jkikstra:
> 
> Could we consider either adopting a more AR6-like structure under `Final Energy|Transportation`, or at least have some aggregated variable corresponding to road transport (and preferably also aviation and maritime/navigation including both domestic and international, cf. #36 )?

I'd also support to have the split of domestic and international aviation under this variable.

PR #188, opened 2024-10-25
Cmip7 Emissions Variables List (v0)

jkikstra:
Implementing the variable list from [here [sheet: IAMC variables]](https://docs.google.com/spreadsheets/d/1HKfl21NX8wohPEHpHlmf5XithknPw8zjLpVe4MiOU9U/edit?usp=sharing)

General context documented under #151 (currently bit outdated).

Should address:

Minimum:
- [x] CDR under "Emissions|*" #145 
    - [x] check consistency with https://github.com/IAMconsortium/common-definitions/pull/172
- [x] High-level (CO2) variables; components Product Use and Fossil Fuel Fires #165 
- [x] Burning variables #167 and #64
    - [x] Update descriptions
- [x] Add "Gross Emissions|CO2|AFOLU" #151 
- [x] Supply + Demand = Total #173
- [ ] merge check: are "level" 1/2/3 components defined consistently between 'component' and the actual variables?

Optional / can leave for future versions:
- [ ] #185 
- [ ] #166
- [ ] #144
- [ ] #187
- [ ] #186 
- [ ] #184 (now domestic aviation would be calculated by adding/subtracting other variables [see here [sheet: IAMC variables]](https://docs.google.com/spreadsheets/d/1HKfl21NX8wohPEHpHlmf5XithknPw8zjLpVe4MiOU9U/edit?usp=sharing))
- [ ] add tier levels (#179)
- [ ] create a 'diff' with the AR6 database template
- [ ] update synfuels guidelines (#49)
- [x] national GHG inventories (#146)
- [ ] consider changing the hierarchy by placing Fires outside of AFOLU (#200)
- [ ] clarify HWP reporting (PR/issue tbd)
- [ ] provide clarity on which sector-species combinations are expected to be zero / not required [e.g. BC|Agriculture] (#202)
- [ ] consider adding CDR methods not only in top-level descriptions, but also to lower-level variable descriptions (PR/issue tbd)
- [ ] reconsider reporting unit for NO2 (NOx / NO2 - NOx may be a bit ambiguous)  (PR/issue tbd)

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -43,7 +43,7 @@
       For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
       - Emissions|{Level-3 Species}|AFOLU|Agriculture
-      - Emissions|{Level-3 Species}|AFOLU|Biomass Burning
+      - Emissions|{Level-3 Species}|AFOLU|Agricultural Waste Burning
       - Emissions|{Level-3 Species}|AFOLU|Land
 - Gross Emissions|CO2|AFOLU:
   description: Gross emissions of carbon dioxide (CO2)  from agriculture, forestry
@@ -56,28 +56,36 @@
 
 - Emissions|{Level-3 Species}|AFOLU|Agriculture:
     description: Emissions of {Level-3 Species} from the agriculture sector
+      (IPCC 1996 category 4; IPCC 2006 category 3A and 3C except 3C1)
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Agriculture|Livestock:
     description: Emissions of {Level-3 Species} from livestock in the agriculture sector
+      (IPCC 1996 category 4A and 4B; IPCC 2006 category 3A)
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Agriculture|Livestock|Enteric Fermentation:
     description: Emissions of {Level-3 Species} from enteric fermentation of livestock
+      (IPCC 1996 category 4A; IPCC 2006 category 3A1)
       in the agriculture sector
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Agriculture|Livestock|Manure Management:
     description: Emissions of {Level-3 Species} from processes involving
       manure (waste) management in Livestock in the agriculture sector
+      (IPCC 1996 category 4B; IPCC 2006 category 3A2)
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Agriculture|Rice:
     description: Emissions of {Level-3 Species} from rice cultivation
+      (IPCC 1996 category 4C; IPCC 2006 category 3C7)
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Agriculture|Managed Soils:
     description: Emissions of {Level-3 Species} from soil management practices
       in the agriculture sector
+      (IPCC 1996 category 4D; IPCC 2006 category 3C4 and 3C5)
     unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Biomass Burning:
+- Emissions|{Level-3 Species}|AFOLU|Agricultural Waste Burning:
     description: Emissions from agricultural waste burning
+      (IPCC 1996 category 4F; partially IPCC 2006 category 3C1)
     unit: "{Level-3 Species}"
+    notes: Also sometimes known as 'biomass burning' or 'open agricultural burning'
 - Emissions|{Level-3 Species}|AFOLU|Land:
     description: Emissions of {Level-3 Species} from forestry and other land use and 
       land use change. Removals in this variable include agroforestry, re/afforestation, 
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -43,12 +43,6 @@
       see "Emissions|*". Emissions reported in this category represent 
       fluxes from the land pool to the atmosphere.
   unit: Mt CO2/yr
-- Gross Emissions|CO2|AFOLU|Deforestation:
-  description: Gross emissions of carbon dioxide (CO2) from deforestation
-- Gross Emissions|CO2|AFOLU|Other Land:
-  description: Gross emissions of carbon dioxide (CO2) from other land-use change
-- Gross Emissions|CO2|AFOLU|Forest Management:
-  description: Gross emissions of carbon dioxide (CO2) from forest management
 
 - Emissions|{Level-3 Species}|AFOLU|Agriculture:
     description: Emissions of {Level-3 Species} from the agriculture sector
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -42,15 +42,15 @@
     notes: Emissions reported in this category represent fluxes to/from the atmosphere from/to the land pool. 
       For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
-      - Emissions|{Level-3 Species}|AFOLU|Agriculture
-      - Emissions|{Level-3 Species}|AFOLU|Agricultural Waste Burning
-      - Emissions|{Level-3 Species}|AFOLU|Land
+      - Emissions|{Level-2 Species}|AFOLU|Agriculture
+      - Emissions|{Level-2 Species}|AFOLU|Agricultural Waste Burning
+      - Emissions|{Level-2 Species}|AFOLU|Land
 - Gross Emissions|CO2|AFOLU:
   description: Gross emissions of carbon dioxide (CO2)  from agriculture, forestry
       and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3).
       This is emissions sources only. All removals (for example, 
       removals from biochar) are not included here. For net emissions, 
-      see "Emissions|*". Emissions reported in this category represent 
+      see "Emissions|CO2|AFOLU". Emissions reported in this category represent 
       fluxes from the land pool to the atmosphere.
   unit: Mt CO2/yr
 
@@ -102,7 +102,7 @@
 - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change:
     description: Emissions and removals from land use and land-use change,
       including emissions from deforestation and conversion from all other natural land,
-      as well as removals from the atmosphere through agroforestry, re/afforestation
+      as well as removals from the atmosphere through agroforestry or re/afforestation
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change|Wetlands:
     description: Emissions from land that is covered or saturated by water for
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -1,6 +1,9 @@
 # Emissions species are grouped into three levels of detail for the reporting.
 # Level-1 reports aggregate emissions, Level-3 has the most detailed disaggregation.
 # Gross emissions are only defined for CO2.
+# Emissions reporting guidance follows IPCC guidance, referring to 
+# 2006 categories: https://www.ipcc-nggip.iges.or.jp/public/2019rf/pdf/1_Volume1/19R_V1_Ch08_Reporting_Guidance.pdf
+# 1996 categories: https://github.com/primap-community/climate_categories/blob/main/climate_categories/data/IPCC1996.yaml
 
 - Emissions|{Level-1 Species}:
     description: Emissions of {Level-1 Species}, net of emissions to the atmosphere and the removal of {Level-1 Species} from the atmosphere.
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -294,8 +294,7 @@
       technologies that are not directly linked 
       to an emissions source, including direct air carbon capture and storage (DACCS) or enhanced 
       weathering (EW), as well as durable wood products in building elements, mineral products, 
-      ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture, 
-      and other methods
+      ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture
     unit: "{Level-3 Species}"
     note: This timeseries should be reported as negative values so that subcategories of
       emissions add up to net emissions `Emissions|{Level-3 Species}`.
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -12,7 +12,7 @@
       contributions. For positive contributions only, see 
       "Gross Emissions|*").
     unit: "{Level-1 Species}"
-    note: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
+    notes: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
       - Emissions|{Level-1 Species}|AFOLU
       - Emissions|{Level-1 Species}|Energy
@@ -39,7 +39,7 @@
       use CCS such as soil carbon sequestration, biochar, forestry, and re/afforestation and 
       agroforestry. 
     unit: "{Level-2 Species}"
-    note: Emissions reported in this category represent fluxes to/from the atmosphere from/to the land pool. 
+    notes: Emissions reported in this category represent fluxes to/from the atmosphere from/to the land pool. 
       For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
       - Emissions|{Level-3 Species}|AFOLU|Agriculture
@@ -82,7 +82,7 @@
     description: Emissions of {Level-3 Species} from forestry and other land use and 
       land use change
     unit: "{Level-3 Species}"
-    note: Emissions from land use change are reported under the land use type
+    notes: Emissions from land use change are reported under the land use type
       that the land is converted to. For instance, deforestation for the creation
       of new cropland is reported under "Emissions|*|AFOLU|Land|Cropland".
     components:
@@ -96,7 +96,7 @@
 - Emissions|{Level-3 Species}|AFOLU|Land|Forest Land:
     description: Emissions and removals from lands with woody vegetation
       (partially IPCC 1996 category 5A/5B/5D; IPCC 2006 category 3B1) 
-    note: It also includes systems with vegetation that currently fall below, 
+    notes: It also includes systems with vegetation that currently fall below, 
       but are expected to later exceed, the threshold values used by a country to
       define the forest land category.
     unit: "{Level-3 Species}"
@@ -105,7 +105,7 @@
       fields, and agro-forestry systems where vegetation falls
       below the thresholds used for the forest land category
       (no clear IPCC 1996 category; IPCC 2006 category 3B2)
-    note: This does not include methane emissions from rice cultivation
+    notes: This does not include methane emissions from rice cultivation
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Land|Grassland:
     description: Emissions and removals from rangelands and pasture land
@@ -113,7 +113,7 @@
       woody vegetation that fall below the threshold values used
       in the forest land category and are not expected to exceed
       them, without human intervention (partially IPCC 1996 category 4D/4E/5A/5B/5C/5D; IPCC 2006 category 3B3).
-    note: The category also includes all grassland from wild lands to recreational 
+    notes: The category also includes all grassland from wild lands to recreational 
       areas as well as agricultural and silvi-pastural systems, subdivided
       into managed and unmanaged, consistent with national
       definitions.
@@ -123,7 +123,7 @@
       all or part of the year (e.g., peatland) and that does not fall
       into the forest land, cropland, grassland or settlements
       categories (partially IPCC 1996 category 4D/5A/5B/5E; IPCC 2006 category 3B4) 
-    note: It includes reservoirs as a managed sub-division and 
+    notes: It includes reservoirs as a managed sub-division and 
       natural rivers and lakes as unmanaged sub-divisions
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Land|Settlements:
@@ -157,7 +157,7 @@
       including emissions from international bunker fuels. This is the sum of sources and sinks, and 
       includes negative emisisons in these sectors, for instance from bioenergy with CCS (BECCS)
     unit: "{Level-3 Species}"
-    note: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
+    notes: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
       - Emissions|{Level-3 Species}|Energy
       - Emissions|{Level-3 Species}|Industrial Processes
@@ -178,7 +178,7 @@
       emissions from carbon capture and removal using BECCS and other forms of CCS on energy supply 
       and demand for instance for Oil, Gas, and Coal, as well as bio-oil storage
     unit: "{Level-2 Species}"
-    note: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
+    notes: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
       - Emissions|{Level-3 Species}|Energy|Supply
       - Emissions|{Level-3 Species}|Energy|Demand
@@ -314,7 +314,7 @@
       and sinks, including negative emissions such as those from cement carbonation
       processes and other forms of capture and storage on industrial processes
     unit: "{Level-3 Species}"
-    note: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
+    notes: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
 - Gross Emissions|CO2|Industrial Processes:
     description: Gross emissions of carbon dioxide (CO2) from industrial processes
       (IPCC categories 1A2, 1A5, 2A, 2B, 2C, 2E), not accounting for negative emissions 
@@ -368,6 +368,6 @@
       weathering (EW), as well as durable wood products in building elements, mineral products, 
       ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture
     unit: "{Level-3 Species}"
-    note: This timeseries should be reported as negative values so that subcategories of
+    notes: This timeseries should be reported as negative values so that subcategories of
       emissions add up to net emissions `Emissions|{Level-3 Species}`. For a breakdown of carbon 
       removals, see the "Carbon Removal|*" tree
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -21,8 +21,10 @@
       - Emissions|{Level-1 Species}|Product Use
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
-    description: Gross emissions of carbon dioxide (CO2), not accounting for 
-      emissions removed from the atmosphere
+    description: Gross emissions of carbon dioxide (CO2). 
+      This is emissions sources only. All removals (for example, 
+      removals from BECCS or direct air carbon capture and storage (DACCS)) 
+      are not included here. For net emissions, see "Emissions|*".
     unit: Mt CO2/yr
 
 - Emissions|{Level-2 Species}|AFOLU:
@@ -34,10 +36,12 @@
       category represent fluxes to/from the atmosphere from/to the land pool.
     unit: "{Level-2 Species}"
 - Gross Emissions|CO2|AFOLU:
-  description: Gross emissions of carbon dioxide (CO2), i.e. the sum of all positive emissions 
-      (excluding all negative emissions), from agriculture, forestry
-      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3). Emissions reported in this 
-      category represent fluxes from the land pool to the atmosphere.
+  description: Gross emissions of carbon dioxide (CO2)  from agriculture, forestry
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3).
+      This is emissions sources only. All removals (for example, 
+      removals from biochar) are not included here. For net emissions, 
+      see "Emissions|*". Emissions reported in this category represent 
+      fluxes from the land pool to the atmosphere.
   unit: Mt CO2/yr
 - Gross Emissions|CO2|AFOLU|Deforestation:
   description: Gross emissions of carbon dioxide (CO2) from deforestation
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -96,7 +96,7 @@
     description: Gross emissions of carbon dioxide (CO2) from energy use in supply and demand sectors
       (IPCC category 1A, 1B) and from industrial processes (IPCC categories 1A2, 1A5, 2A, 2B, 2C, 2E),
       including emissions from international bunker fuels, not accounting for 
-      emissions removed from the atmosphere
+      removals from the atmosphere
     unit: Mt CO2/yr
     components:
       - Gross Emissions|CO2|Energy
@@ -196,7 +196,7 @@
     unit: "{Level-2 Species}"
 - Gross Emissions|CO2|Energy|Demand|Industry:
     description: Gross emissions of carbon dioxide (CO2) from fuel combustion in industry (IPCC category 1A2),
-      not accounting for emissions removed from the atmosphere
+      not accounting for removals from the atmosphere
     unit: Mt CO2/yr
 - Emissions|CO2|Energy|Demand|Industry|{Non-Energy Sector}:
     description: Emissions of carbon dioxide (CO2) from energy demand in the {Non-Energy Sector}
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -30,13 +30,16 @@
 
 - Emissions|{Level-2 Species}|AFOLU:
     description: Emissions of {Level-2 Species} from agriculture, forestry
-      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3), 
-      reporting the sum of sources and sinks including land-use methods that do not 
-      use CCS such as soil carbon sequestration, biochar, forestry, peatland 
-      and wetland restoration, and biomass burial. Emissions reported in this 
-      category represent fluxes to/from the atmosphere from/to the land pool.
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3).
+      In particular, this includes agricultural emissions (mostly non-CO2) 
+      from livestock, rice cultivation and soil management as well as 
+      emissions from land use and land-use change (mostly CO2) such as deforestation, 
+      conversion of non-forest natural land, and drained peatlands. 
+      This is reporting the sum of sources and sinks including land-use methods that do not 
+      use CCS such as soil carbon sequestration, biochar, forestry, and re/afforestation and 
+      agroforestry. 
     unit: "{Level-2 Species}"
-    note: For a breakdown of carbon removals, see the "Carbon Removal|*" tree
+    note: Emissions reported in this category represent fluxes to/from the atmosphere from/to the land pool. For a breakdown of carbon removals, see the "Carbon Removal|*" tree
     components:
       - Emissions|{Level-3 Species}|AFOLU|Agriculture
       - Emissions|{Level-3 Species}|AFOLU|Biomass Burning
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -95,60 +95,37 @@
       that the land is converted to. For instance, deforestation for the creation
       of new cropland is reported under "Emissions|*|AFOLU|Land|Cropland".
     components:
-      - Emissions|{Level-3 Species}|AFOLU|Land|Forest Land
-      - Emissions|{Level-3 Species}|AFOLU|Land|Cropland
-      - Emissions|{Level-3 Species}|AFOLU|Land|Grassland
-      - Emissions|{Level-3 Species}|AFOLU|Land|Wetlands
-      - Emissions|{Level-3 Species}|AFOLU|Land|Settlements
-      - Emissions|{Level-3 Species}|AFOLU|Land|Other Land
+      - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change
       - Emissions|{Level-3 Species}|AFOLU|Land|Fires
-- Emissions|{Level-3 Species}|AFOLU|Land|Forest Land:
-    description: Emissions and removals from lands with woody vegetation
-      (partially IPCC 1996 category 5A/5B/5D; IPCC 2006 category 3B1) 
-    notes: It also includes systems with vegetation that currently fall below, 
-      but are expected to later exceed, the threshold values used by a country to
-      define the forest land category.
+      - Emissions|{Level-3 Species}|AFOLU|Land|Harvested Wood Products
+      - Emissions|{Level-3 Species}|AFOLU|Land|Other
+- Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change:
+    description: Emissions and removals from land use and land-use change,
+      including emissions from deforestation and conversion from all other natural land,
+      as well as removals from the atmosphere through agroforestry, re/afforestation
     unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Cropland:
-    description: Emissions and removals from arable and tillage land, rice
-      fields, and agro-forestry systems where vegetation falls
-      below the thresholds used for the forest land category
-      (no clear IPCC 1996 category; IPCC 2006 category 3B2)
-    notes: This does not include methane emissions from rice cultivation
-    unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Grassland:
-    description: Emissions and removals from rangelands and pasture land
-      that is not considered cropland. It also includes systems with
-      woody vegetation that fall below the threshold values used
-      in the forest land category and are not expected to exceed
-      them, without human intervention (partially IPCC 1996 category 4D/4E/5A/5B/5C/5D; IPCC 2006 category 3B3).
-    notes: The category also includes all grassland from wild lands to recreational 
-      areas as well as agricultural and silvi-pastural systems, subdivided
-      into managed and unmanaged, consistent with national
-      definitions.
-    unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Wetlands:
+- Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change|Wetlands:
     description: Emissions from land that is covered or saturated by water for
       all or part of the year (e.g., peatland) and that does not fall
       into the forest land, cropland, grassland or settlements
       categories (partially IPCC 1996 category 4D/5A/5B/5E; IPCC 2006 category 3B4) 
-    notes: It includes reservoirs as a managed sub-division and 
-      natural rivers and lakes as unmanaged sub-divisions
+    notes: Includes reservoirs as a managed sub-division and 
+      natural rivers and lakes as unmanaged sub-divisions. Includes drained 
+      and rewetted wetlands, does not include natural emissions from intact wetlands.
     unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Settlements:
-    description: Emissions and removals from all developed land, including
-      transportation infrastructure and human settlements of any
-      size, unless they are already included under other
-      categories (partially IPCC 1996 category 5A/5B/5D/5E; IPCC 2006 category 3B5).
+- Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change|Other:
+    description: Emissions and removals from forest land, cropland, grassland,
+      settlements, and other land that cannot be accommodated in other categories
+      (partially IPCC 1996 category 5; IPCC 2006 category 3B except 3B4). 
     unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Other Land:
-    description: Emissions and removals from bare soil, rock, ice, and all
-      unmanaged land areas that do not fall into any of the other
-      land categories (no clear IPCC 1996 category; IPCC 2006 category 3B6). 
-    unit: "{Level-3 Species}"
+    notes: Only includes anthropogenic emissions. Does not include wetlands.
 - Emissions|{Level-3 Species}|AFOLU|Land|Fires:
     description: Emissions from land fires and deforestation & degradation
     unit: "{Level-3 Species}"
+    components:
+      - Emissions|{Level-3 Species}|AFOLU|Land|Fires|Peat Burning
+      - Emissions|{Level-3 Species}|AFOLU|Land|Fires|Forest Burning
+      - Emissions|{Level-3 Species}|AFOLU|Land|Fires|Grassland Burning
 - Emissions|{Level-3 Species}|AFOLU|Land|Fires|Peat Burning:
     description: Emissions from peat fires
     unit: "{Level-3 Species}"
@@ -158,6 +135,15 @@
     unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Land|Fires|Grassland Burning:
     description: Emissions from savanna, grassland, and shrubland fires
+    unit: "{Level-3 Species}"  
+- Emissions|{Level-3 Species}|AFOLU|Land|Harvested Wood Products:
+    description: Emissions and removals from harvested wood products (HWP)
+      (no clear IPCC 1996 category; IPCC 2006 category 3D1)
+    unit: "{Level-3 Species}"
+- Emissions|{Level-3 Species}|AFOLU|Land|Other:
+    description: Emissions and removals from land that cannot be 
+      accommodated in other categories (no clear IPCC 1996 category; 
+      IPCC 2006 category 3D2) 
     unit: "{Level-3 Species}"
 
 - Emissions|{Level-3 Species}|Energy and Industrial Processes:
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -286,7 +286,7 @@
 - Emissions|{Level-2 Species}|Other:
     description: Emissions of {Level-2 Species} from other sources, (IPCC 1996 category 7; IPCC 2006 category 5B; including fossil fuel fires). 
       Only use this category exceptionally, for any categories than cannot be accommodated in the categories described above. 
-      Unless specified otherwise, this will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
+      notes: This will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
     unit: "{Level-2 Species}"
 
 - Emissions|{Level-3 Species}|Other Capture and Removal:
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -287,7 +287,7 @@
     description: Emissions of {Level-2 Species} from other sources (IPCC 1996 category 7;
       IPCC 2006 category 5B) including fossil fuel fires and any sources or sinks that
       cannot be accommodated in other categories
-      notes: This will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
+      notes: This will be treated as a flux from the fossil reservoir to the atmosphere in climate models.
     unit: "{Level-2 Species}"
 
 - Emissions|{Level-3 Species}|Other Capture and Removal:
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -29,6 +29,19 @@
       and wetland restoration, and biomass burial. Emissions reported in this 
       category represent fluxes to/from the atmosphere from/to the land pool.
     unit: "{Level-2 Species}"
+- Gross Emissions|CO2|AFOLU:
+  description: Gross emissions of carbon dioxide (CO2), i.e. the sum of all positive emissions 
+      (excluding all negative emissions), from agriculture, forestry
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3). Emissions reported in this 
+      category represent fluxes from the land pool to the atmosphere.
+  unit: Mt CO2/yr
+- Gross Emissions|CO2|AFOLU|Deforestation:
+  description: Gross emissions of carbon dioxide (CO2) from deforestation
+- Gross Emissions|CO2|AFOLU|Other Land:
+  description: Gross emissions of carbon dioxide (CO2) from other land-use change
+- Gross Emissions|CO2|AFOLU|Forest Management:
+  description: Gross emissions of carbon dioxide (CO2) from forest management
+
 - Emissions|{Level-3 Species}|AFOLU|Agriculture:
     description: Emissions of {Level-3 Species} from the agriculture sector
     unit: "{Level-3 Species}"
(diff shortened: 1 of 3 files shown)


jkikstra:
@danielhuppmann @phackstock apologies for not fully finishing this today. But probably it's already clear where this should go (just the burning variable definitions are left to do for a v0, I think). 

So might be ready for a first review?

Feel free to change things, I don't have anything changed locally right now.

definitions/variable/emissions/emissions.yaml
```diff
@@ -3,24 +3,27 @@
 # Gross emissions are only defined for CO2.
 
 - Emissions|{Level-1 Species}:
-    description: Emissions of {Level-1 Species}
+    description: Emissions of {Level-1 Species}, net of emissions to the atmosphere and the removal of {Level-1 Species} from the atmosphere.
```
danielhuppmann, line 6: I don't think that the description "Emissions of CO2, net of emissions to the atmosphere..." makes a lot of sense

definitions/variable/emissions/emissions.yaml
```diff
@@ -3,24 +3,27 @@
...
       - Emissions|{Level-1 Species}|Waste
-      - Emissions|{Level-1 Species}|Capture and Removal
+      - Emissions|{Level-1 Species}|Other Capture and Removal
+      - Emissions|{Level-1 Species}|Product Use
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
-    description: Gross emissions of carbon dioxide (CO2), not accounting for negative
-      emissions from bioenergy with CCS (BECCS) or agriculture, forestry
-      and other land use (AFOLU)
+    description: Gross emissions of carbon dioxide (CO2), not accounting for
```
danielhuppmann, line 17: Why remove the clear reference to BECCS and negative emissions? Also it's not clear what "not accounting" means in this context...

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,28 +46,33 @@
     description: Emissions of {Level-3 Species} from soil management practices
       in the agriculture sector
     unit: "{Level-3 Species}"
+- Emissions|*|AFOLU|Biomass Burning:
```
danielhuppmann, line 49:
```suggestion
- Emissions|{Level-3 Species}|AFOLU|Biomass Burning:
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -212,20 +226,21 @@
...
-      (IPCC categories 2A, B, C, E), net of negative emissions using CCS
+      (IPCC categories 1A2, 1A5, 2A, 2B, 2C, 2E), net of negative emissions 
+      using CCS such as from cement production
     unit: "{Level-3 Species}"
 - Gross Emissions|CO2|Industrial Processes:
     description: Gross emissions of carbon dioxide (CO2) from industrial processes
-      (IPCC categories 2A, B, C, E), not accounting for negative emissions from
-      bioenergy with CCS (BECCS) in these sectors
+      (IPCC categories 1A2, 1A5, 2A, 2B, 2C, 2E), not accounting for negative emissions 
+      from carbon removal processes such as from cement production
```
danielhuppmann, line 235: Can cement production have negative emissions?

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,28 +46,33 @@
...
     unit: "{Level-3 Species}"
+- Emissions|*|AFOLU|Biomass Burning:
+    description: Agricultural Waste Burning (DESCRIPTION TO BE UPDATED)
+    unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Land:
     description: Emissions of {Level-3 Species} from forestry and other land use
     unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Wetlands:
-    description: Emissions of {Level-3 Species} from managed peatlands (drained and rewetted)
+- Emissions|{Level-3 Species}|AFOLU|Land|Peat Burning:
```
danielhuppmann, line 55: Is "Peat Burning" really the same as emissions from wetlands? Or do you want to explicitly remove emissions from wetlands?

danielhuppmann (commented):
Thanks @jkikstra! A few clarifying questions and corrections inline.

The main issue, for me, is that that the formulations "net of" or "not accounting for" is inherently ambiguous (should the activity be not counted or the emissions from that activity?)...

I suggest to revert to shorter variable description, and the add a "notes" attribute with a clear guide on how to compute it, like this example
https://github.com/IAMconsortium/common-definitions/blob/88ad6f358cf340211d927cbd4ff439b3305e5b33/definitions/variable/energy/secondary-energy-electricity.yaml#L17

tmkhsgw:
Hi, thank you for your great work! I just let you know Land group would like to have the following variables as gross emissions subcategories for AFOLU (as tier3). Units should be Mt CO2/yr. Thank you.

Gross Emissions|CO2|AFOLU|Deforestation:
description: Gross emissions of CO2 from deforestation
Gross Emissions|CO2|AFOLU|Other Land:
description: Gross emissions of CO2 from forest management
Gross Emissions|CO2|AFOLU|Forest Management:
description: Gross emissions of CO2 from other land-use change

definitions/variable/emissions/emissions.yaml
znichollscr: Yep there's a carbonation sink (see e.g. https://robbieandrew.github.io/GCB2023/PNG/s41_2023_Carbonation.png)

definitions/variable/emissions/emissions.yaml
```diff
@@ -251,18 +266,17 @@
...
     unit: "{Level-2 Species}"
 
-- Emissions|{Level-2 Species}|Fossil Fuel Fires:
-    description: Emissions of {Level-2 Species} from fossil fuel fires (IPCC category 7)
-    unit: "{Level-2 Species}"
-
 - Emissions|{Level-2 Species}|Other:
-    description: Emissions of {Level-2 Species} from other sources
+    description: Emissions of {Level-2 Species} from other sources, (IPCC category 7, including fossil fuel fires)
     unit: "{Level-2 Species}"
```
znichollscr, line 271: This variable is a little ambiguous. In any processing, I would just assume that it is a flux from some fossil reservoir to the atmosphere but that could be wrong, depending on what it actually is.

definitions/variable/emissions/emissions.yaml
```diff
@@ -251,18 +266,17 @@
...
-    description: Capture and removal of atmospheric {Level-3 Species} using net-negative
-      technologies that are not directly linked to an emissions source,
-      e.g., direct air capture (DAC) or enhanced weathering
+- Emissions|{Level-3 Species}|Other Capture and Removal:
+    description: Capture and removal of atmospheric {Level-3 Species} using other net-negative 
+      technologies that are not reported under other variables as they are not directly linked 
+      to an emissions source, including direct air carbon capture and storage (DACCS) or enhanced 
+      weathering (EW), as well as durable wood products in building elements, mineral products, 
+      ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture, 
+      and other methods
```
znichollscr, line 279: I think it would be ideal to make this slightly clearer in terms of where the carbon is going, specifically does it go into some permanent storage, the land pool or the ocean (or somewhere else). Maybe that means we just need to add components?

definitions/variable/emissions/emissions.yaml
```diff
@@ -251,18 +266,17 @@
     description: Emissions of {Level-2 Species} from waste (IPCC category 6)
```
znichollscr, line 266:
```suggestion
    description: Emissions of {Level-2 Species} from waste (IPCC 1996 category 6)
```

A potentially handy cross-reference https://github.com/primap-community/climate_categories/blob/main/climate_categories/data/IPCC1996.yaml. Might be worth pulling up the guidance here that "CO2 from organic waste handling and decay should not be included", i.e. this category is just for emissions from fossil-based sources. (Having said that, if we're not doing this perfectly, let's not spend too much time on it.)

definitions/variable/emissions/emissions.yaml
znichollscr: It can be helpful to be explicit that we're talking about net emissions here. However, if the default assumption in the common definitions is that "Emissions" always means net, and we deal with gross emissions and removals some other way, then it could be more confusing than helpful to repeat the definition here (because it implies that the convention isn't actually a convention).

definitions/variable/emissions/emissions.yaml
znichollscr: > Why remove the clear reference to BECCS and negative emissions?

I think it is because there are more negative emissions options than just BECCS now. Rather than repeating them all, it's quicker and clearer to just say, "excluding all negative".

definitions/variable/emissions/emissions.yaml
danielhuppmann: When I read "A, net of B", I think it means "A minus B" - that's what I meant with "doesn't make a lot of sense"...

definitions/variable/emissions/emissions.yaml
```diff
@@ -3,24 +3,27 @@
...
     unit: "{Level-1 Species}"
     components:
       - Emissions|{Level-1 Species}|AFOLU
       - Emissions|{Level-1 Species}|Energy
       - Emissions|{Level-1 Species}|Industrial Processes
       - Emissions|{Level-1 Species}|Waste
-      - Emissions|{Level-1 Species}|Capture and Removal
+      - Emissions|{Level-1 Species}|Other Capture and Removal
+      - Emissions|{Level-1 Species}|Product Use
       - Emissions|{Level-1 Species}|Other
```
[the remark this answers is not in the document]
danielhuppmann: Yes, it is. Defining how the components should be understood and how they should be aggregated is obviously an important part of a definition.

definitions/variable/emissions/emissions.yaml
```diff
@@ -73,22 +81,28 @@
...
+    description: Emissions of carbon dioxide (CO2) from energy use, including 
+      fugitive emissions from fuels (IPCC category 1A (except manufacturing 1A2 
+      and other unspecified 1A5), 1B), reporting the net of energy emissions and 
+      carbon capture and removal using BECCS and other forms of CCS on energy forms 
+      such Natural Gas, Oil, and Coal, as well as bio-oil storage
     unit: "{Level-2 Species}"
+    components:
+      - Emissions|{Level-3 Species}|Energy|Supply
+      - Emissions|{Level-3 Species}|Energy|Demand
 - Gross Emissions|CO2|Energy:
```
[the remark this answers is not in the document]
danielhuppmann: This was discussed at length in earlier projects.

definitions/variable/emissions/emissions.yaml
danielhuppmann: Thanks!

definitions/variable/emissions/emissions.yaml
danielhuppmann: There's a detailed "Carbon Removal|..." variable tree for that. This variable just makes the Emissions-tree add up.

definitions/variable/emissions/emissions.yaml
jkikstra: Ahh that's very useful to have the IPCC categories in that overview too. 

Should be made clear which guidance is used, because 1996 and 2006 guidance are different. 
I noted this in #186; as we perhaps want to make a separate label for it?

definitions/variable/emissions/emissions.yaml
znichollscr: Ah I see, I agree. What do you think of the below

```suggestion
    description: Emissions of {Level-1 Species} to the atmosphere. For the avoidance of doubt, these emissions are net emissions (i.e. include both positive (emission) and negative (removal) contributions. For positive contributions only, see "Gross Emissions|*").
```

definitions/variable/emissions/emissions.yaml
znichollscr: Ok. Given that negative emissions is not just BECCS, I do think it needs to be somehow clear that all removals are excluded? Something like the below?

```suggestion
    description: Gross emissions of carbon dioxide (CO2). This is emissions only. All removals (for example, removals from BECCS or direct air capture (DAC)) are not included here.
```

definitions/variable/emissions/emissions.yaml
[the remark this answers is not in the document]
znichollscr: Cool thanks. (For some reason I can't resolve this thread, but I don't have any further questions)

definitions/variable/emissions/emissions.yaml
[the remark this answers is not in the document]
znichollscr: Good to know, thanks

definitions/variable/emissions/emissions.yaml
znichollscr: Ok. @jkikstra this means we'll have to be a bit more careful than we have previously in how we get the input variables for simple climate models. I think we'll basically have to use the "Emissions|*" tree for most things, but then jump into the "Carbon Removal|*" (and maybe the "Gross Emissions|*") trees to work out the details of where the fluxes actually should go.

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,28 +46,33 @@
...
     description: Emissions and removals of {Level-3 Species} from rangelands and pasture land
-      (IPCC category 3B3)
-    unit: "{Level-3 Species}"
-- Emissions|{Level-3 Species}|AFOLU|Land|Other Land Burning:
-    description: Emissions and removals of {Level-3 Species} from any other categories
-      of land burning (e.g. peatlands)
+      (IPCC category 3B3)  (DESCRIPTION TO BE UPDATED)
     unit: "{Level-3 Species}"
 
+- Gross Emissions|CO2|AFOLU:
```
danielhuppmann, line 67: Thank you for adding this, but can you please move this to just below `Emissions|{Level-2 Species}|AFOLU` and then have an empty line for consistency of the ordering?

definitions/variable/emissions/emissions.yaml
jkikstra: In part that is because the IPCC category also is "Only use this category exceptionally, for any categories than cannot be accommodated in the categories described above. Include a reference to where a detailed explanation of the category can be found."

Maybe I'll just add that.

jkikstra: How's this d4fad19b34d41528f5db215191d91a999f9566ea

definitions/variable/emissions/emissions.yaml
jkikstra: Agreed - so are we OK with keeping this as is? 
This is just meant just such that people put CDR methods under the same "Emissions|CO2|*" variable. 
At the moment I don't think we need to specify these pools, but if you think it would be good to have this here in addition, then maybe you can create an issue to see if we want to add it later. But I think it would be duplicating the Carbon Removal variable too much.

definitions/variable/emissions/emissions.yaml
jkikstra: Added in d4fad19b34d41528f5db215191d91a999f9566ea

definitions/variable/emissions/emissions.yaml
```diff
@@ -3,24 +3,27 @@
...
+      emissions removed from the atmosphere
     unit: Mt CO2/yr
 
 - Emissions|{Level-2 Species}|AFOLU:
     description: Emissions of {Level-2 Species} from agriculture, forestry
-      and other land use (IPCC category 3)
+      and other land use (IPCC category 3), reporting the net of emissions 
+      including land-use methods that do not use CCS such as soil carbon 
+      sequestration, biochar, forestry, peatland and wetland restoration, 
+      and biomass burial
```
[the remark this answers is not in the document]
jkikstra: I see that this can be useful clarification, both for the reporting and then for use in climate models afterwards.
Added the note as suggested in 28fcd7a7fe6392884a3d06179423660367a946ce

definitions/variable/emissions/emissions.yaml
```diff
@@ -3,24 +3,27 @@
...
-      - Emissions|{Level-1 Species}|Capture and Removal
+      - Emissions|{Level-1 Species}|Other Capture and Removal
+      - Emissions|{Level-1 Species}|Product Use
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
-    description: Gross emissions of carbon dioxide (CO2), not accounting for negative
-      emissions from bioenergy with CCS (BECCS) or agriculture, forestry
-      and other land use (AFOLU)
+    description: Gross emissions of carbon dioxide (CO2), not accounting for 
+      emissions removed from the atmosphere
```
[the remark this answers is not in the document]
jkikstra: Nice - somehow the web view here isn't working for me to commit this, but change to this.

jkikstra:
> Hi, thank you for your great work! I just let you know Land group would like to have the following variables as gross emissions subcategories for AFOLU (as tier3). Units should be Mt CO2/yr. Thank you.
> 
> Gross Emissions|CO2|AFOLU|Deforestation: description: Gross emissions of CO2 from deforestation Gross Emissions|CO2|AFOLU|Other Land: description: Gross emissions of CO2 from forest management Gross Emissions|CO2|AFOLU|Forest Management: description: Gross emissions of CO2 from other land-use change

Added (with minor corrections) here: 3dc5809c1bdb93e018c12092b379b586ea719192

definitions/variable/emissions/emissions.yaml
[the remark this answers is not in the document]
jkikstra: Implemented in 3dc5809c1bdb93e018c12092b379b586ea719192

definitions/variable/emissions/emissions.yaml
jkikstra: 418620d2ee025365922b15204cc4fb9d978869db

definitions/variable/emissions/emissions.yaml
jkikstra: Sounds good to me. How about adding that to the top-level, while changing 
"net of emissions"
to 
"the sum of sources and sinks" everywhere else?

definitions/variable/emissions/emissions.yaml
jkikstra: Done in 821a8d0211a7e1c778e5dcf71a3b5f90fe8ad81b

definitions/variable/emissions/emissions.yaml
jkikstra: In 4ab62596c089f3981e2413ada6b137d86eea18a9

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,27 +1,55 @@
...
+      removals from biochar) are not included here. For net emissions, 
+      see "Emissions|*". Emissions reported in this category represent 
+      fluxes from the land pool to the atmosphere.
+  unit: Mt CO2/yr
+- Gross Emissions|CO2|AFOLU|Deforestation:
+  description: Gross emissions of carbon dioxide (CO2) from deforestation
+- Gross Emissions|CO2|AFOLU|Other Land:
+  description: Gross emissions of carbon dioxide (CO2) from other land-use change
+- Gross Emissions|CO2|AFOLU|Forest Management:
+  description: Gross emissions of carbon dioxide (CO2) from forest management
```
danielhuppmann, line 51: There are no corresponding net-emissions variables for deforestation, other-land and forest-management. I suggest to remove these variables for now and then have a follow-up PR and coordinate with the AFOLU team.

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +284,24 @@
...
+      Unless specified otherwise, this will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
     unit: "{Level-2 Species}"
 
-- Emissions|{Level-3 Species}|Capture and Removal:
-    description: Capture and removal of atmospheric {Level-3 Species} using net-negative
-      technologies that are not directly linked to an emissions source,
-      e.g., direct air capture (DAC) or enhanced weathering
+- Emissions|{Level-3 Species}|Other Capture and Removal:
+    description: Capture and removal of atmospheric {Level-3 Species} using other net-negative 
+      technologies that are not reported under other variables as they are not directly linked
```
danielhuppmann, line 300: Removing this extra qualifier improves readibility.
```suggestion
      technologies that are not directly linked 
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +284,24 @@
...
-    description: Capture and removal of atmospheric {Level-3 Species} using net-negative
-      technologies that are not directly linked to an emissions source,
-      e.g., direct air capture (DAC) or enhanced weathering
+- Emissions|{Level-3 Species}|Other Capture and Removal:
+    description: Capture and removal of atmospheric {Level-3 Species} using other net-negative 
+      technologies that are not reported under other variables as they are not directly linked 
+      to an emissions source, including direct air carbon capture and storage (DACCS) or enhanced 
+      weathering (EW), as well as durable wood products in building elements, mineral products, 
+      ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture, 
+      and other methods
```
danielhuppmann, line 304: This half-sentence starts with "including", so no need to again add "and other methods"
```suggestion
      ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +284,24 @@
...
+    description: Emissions of {Level-2 Species} from waste (IPCC 1996 category 6; IPCC 2006 category 4). 
+      Any emissions from fossil-based products (incineration or CO2 decomposition) should be accounted for here.
+      CO2 from organic waste handling and decay should not be included.
     unit: "{Level-2 Species}"
 
 - Emissions|{Level-2 Species}|Other:
-    description: Emissions of {Level-2 Species} from other sources
+    description: Emissions of {Level-2 Species} from other sources, (IPCC 1996 category 7; IPCC 2006 category 5B; including fossil fuel fires). 
+      Only use this category exceptionally, for any categories than cannot be accommodated in the categories described above. 
+      Unless specified otherwise, this will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
```
danielhuppmann, line 295: There is no way to specify this in a scenario-submission, so I suggest to make it clear that this will be assigned as a fossil flux in climate models.
```suggestion
      notes: This will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +284,24 @@
...
-    description: Emissions of {Level-2 Species} from fossil fuel fires (IPCC category 7)
+    description: Emissions of {Level-2 Species} from waste (IPCC 1996 category 6; IPCC 2006 category 4). 
+      Any emissions from fossil-based products (incineration or CO2 decomposition) should be accounted for here.
+      CO2 from organic waste handling and decay should not be included.
     unit: "{Level-2 Species}"
 
 - Emissions|{Level-2 Species}|Other:
-    description: Emissions of {Level-2 Species} from other sources
+    description: Emissions of {Level-2 Species} from other sources, (IPCC 1996 category 7; IPCC 2006 category 5B; including fossil fuel fires). 
+      Only use this category exceptionally, for any categories than cannot be accommodated in the categories described above.
```
danielhuppmann, line 294: Why is "including fossil fuels" inside the brackets? And why should fossil fuel fires only be reported exceptionally?

Suggest to simplify.
```suggestion
    description: Emissions of {Level-2 Species} from other sources (IPCC 1996 category 7;
      IPCC 2006 category 5B) including fossil fuel fires and any sources or sinks that
      cannot be accommodated in other categories
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +284,24 @@
...
 
 - Emissions|{Level-2 Species}|Waste:
-    description: Emissions of {Level-2 Species} from waste (IPCC category 6)
-    unit: "{Level-2 Species}"
-
-- Emissions|{Level-2 Species}|Fossil Fuel Fires:
-    description: Emissions of {Level-2 Species} from fossil fuel fires (IPCC category 7)
+    description: Emissions of {Level-2 Species} from waste (IPCC 1996 category 6; IPCC 2006 category 4). 
+      Any emissions from fossil-based products (incineration or CO2 decomposition) should be accounted for here.
+      CO2 from organic waste handling and decay should not be included.
```
danielhuppmann, line 289:
```suggestion
    description: Emissions of {Level-2 Species} from waste of fossil-based products
      through incineration or decomposition (IPCC 1996 category 6; IPCC 2006 category 4), 
      not including emissions from organic waste handling and decay
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,27 +1,55 @@
...
+      - Emissions|{Level-1 Species}|Product Use
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
-    description: Gross emissions of carbon dioxide (CO2), not accounting for negative
-      emissions from bioenergy with CCS (BECCS) or agriculture, forestry
-      and other land use (AFOLU)
+    description: Gross emissions of carbon dioxide (CO2). 
+      This is emissions sources only. All removals (for example, 
+      removals from BECCS or direct air carbon capture and storage (DACCS)) 
+      are not included here. For net emissions, see "Emissions|*".
```
danielhuppmann, line 27:
```suggestion
    description:  Gross emissions of carbon dioxide (CO2), not accounting for
      negative  emissions from bioenergy with CCS (BECCS), direct air carbon capture
      or agriculture, forestry and other land use (AFOLU)
    notes: For net emissions (accounting for negative emissions), see "Emissions|CO2".
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +284,24 @@
...
+    description: Emissions of {Level-2 Species} from other sources, (IPCC 1996 category 7; IPCC 2006 category 5B; including fossil fuel fires). 
+      Only use this category exceptionally, for any categories than cannot be accommodated in the categories described above. 
+      Unless specified otherwise, this will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
     unit: "{Level-2 Species}"
 
-- Emissions|{Level-3 Species}|Capture and Removal:
-    description: Capture and removal of atmospheric {Level-3 Species} using net-negative
-      technologies that are not directly linked to an emissions source,
-      e.g., direct air capture (DAC) or enhanced weathering
+- Emissions|{Level-3 Species}|Other Capture and Removal:
```
danielhuppmann, line 298: How about renaming that to "Atmospheric Capture and Removal" to be more specific?

jkikstra: But the reason for "Other" is to make clear that it is not supposed to also include the CDR already under "Energy"/"Industrial Processes"/"AFOLU".

So would you like "Other Atmospheric Capture and Removal"?

danielhuppmann: Ok, never mind my suggestion…

definitions/variable/emissions/emissions.yaml
jkikstra: OK now in https://github.com/IAMconsortium/common-definitions/pull/199

definitions/variable/emissions/emissions.yaml
jkikstra: Sure, I thought for me it would be helpful (this is how I would explain it to others) but fine.

definitions/variable/emissions/emissions.yaml
jkikstra: Fossil fuel fires are reported in CEDS, and thus would have to go somewhere in a mapping to IAMC variables. It doesn't go anywhere else, but to me and Oliver it wasn't immediately clear where it should go. To remove confusion, I think it should be specified that we put it under "Other", not under e.g. "Energy". That's in line with EDGAR and CEDS.

danielhuppmann: Yes, I think it makes sense to put fossil fuel fires here, but then the rest of the description is strange - and I tried to clarify that with the alternative proposal...

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,27 +1,49 @@
...
+      - Emissions|{Level-1 Species}|Other Capture and Removal
+      - Emissions|{Level-1 Species}|Product Use
       - Emissions|{Level-1 Species}|Other
 - Gross Emissions|CO2:
-    description: Gross emissions of carbon dioxide (CO2), not accounting for negative
-      emissions from bioenergy with CCS (BECCS) or agriculture, forestry
-      and other land use (AFOLU)
+    description:  Gross emissions of carbon dioxide (CO2), not accounting for
+      negative  emissions from bioenergy with CCS (BECCS), direct air carbon capture
+      or agriculture, forestry and other land use (AFOLU)
```
znichollscr, line 26:
```suggestion
    description:  Gross emissions of carbon dioxide (CO2), not accounting for
      negative emissions, for example from bioenergy with CCS (BECCS), direct air carbon capture
      or agriculture, forestry and other land use (AFOLU)
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,52 +65,57 @@
...
     unit: "{Level-2 Species}"
+    components:
+      - Emissions|{Level-3 Species}|Energy|Supply
+      - Emissions|{Level-3 Species}|Energy|Demand
 - Gross Emissions|CO2|Energy:
     description: Gross emissions of carbon dioxide (CO2) from energy use,
-      including fugitive emissions from fuels (IPCC category 1A, 1B), not accounting for
-      negative emissions from bioenergy with CCS (BECCS) in these sectors
+      including fugitive emissions from fuels (IPCC category 1A, 1B), not accounting for 
+      emissions removed from the atmosphere
```
znichollscr, line 118:
```suggestion
      removals from the atmosphere
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -104,8 +131,8 @@
       other energy conversion (e.g., refineries, synfuel production, solid fuel processing,
       IPCC category 1Ab, 1Ac), incl. pipeline transportation (IPCC category 1A3ei),
       fugitive emissions from fuels (IPCC category 1B) and emissions from carbon dioxide
-      transport and storage (IPCC category 1C), not accounting for negative emissions
-      from bioenergy with CCS (BECCS) in these sectors
+      transport and storage (IPCC category 1C), not accounting for 
+      emissions removed from the atmosphere
```
znichollscr, line 135:
```suggestion
      removals from the atmosphere
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -123,52 +150,53 @@
     unit: Mt CO2/yr
 - Gross Emissions|CO2|Energy|Supply|Electricity:
     description: Gross emissions of carbon dioxide (CO2) for electricity generation,
-      not accounting for negative emissions from bioenergy with CCS (BECCS) in this sector
+      not accounting for emissions removed from the atmosphere
```
znichollscr, line 153:
```suggestion
      not accounting for removals from the atmosphere
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -123,52 +150,53 @@
...
     unit: Mt CO2/yr
 - Emissions|CO2|Energy|Supply|{Secondary Fuel Level 2}:
     description: Emissions of carbon dioxide (CO2) for production of {Secondary Fuel Level 2}
     unit: Mt CO2/yr
     tier: "{Secondary Fuel Level 2}"
 - Gross Emissions|CO2|Energy|Supply|{Secondary Fuel Level 2}:
     description: Gross emissions of carbon dioxide (CO2) for production of {Secondary Fuel Level 2},
-      not accounting for negative emissions from bioenergy with CCS (BECCS) in this sector
+      not accounting for emissions removed from the atmosphere
```
znichollscr, line 161:
```suggestion
      not accounting for removals from the atmosphere
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -123,52 +150,53 @@
...
     unit: Mt CO2/yr
     tier: "{Secondary Fuel Level 2}"
 - Emissions|CO2|Energy|Supply|Autoproduction:
     description: Emissions of carbon dioxide (CO2) for own-use energy supply
     unit: Mt CO2/yr
 - Gross Emissions|CO2|Energy|Supply|Autoproduction:
     description: Gross emissions of carbon dioxide (CO2) for own-use energy supply,
-      not accounting for negative emissions from bioenergy with CCS (BECCS) in this sector
+      not accounting for emissions removed from the atmosphere
```
znichollscr, line 169:
```suggestion
      not accounting for removals from the atmosphere
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -123,52 +150,53 @@
...
     unit: Mt CO2/yr
 - Emissions|CO2|Energy|Supply|Other:
     description: Emissions of carbon dioxide (CO2) for other categories of energy supply
     unit: Mt CO2/yr
 - Gross Emissions|CO2|Energy|Supply|Other:
     description: Gross emissions of carbon dioxide (CO2) for other categories of energy supply,
-      not accounting for negative emissions from bioenergy with CCS (BECCS) in these sectors
+      not accounting for emissions removed from the atmosphere
```
znichollscr, line 176:
```suggestion
      not accounting for removals from the atmosphere
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -123,52 +150,53 @@
...
+      This is the sum of sources and sinks, including negative emissions such as the demand-side use 
+      of bioenergy with CCS (BECCS)
     unit: "{Level-2 Species}"
 - Gross Emissions|CO2|Energy|Demand:
     description: Gross emissions of carbon dioxide (CO2) from fuel combustion in industry (IPCC category 1A2),
       residential, commercial, institutional sectors and agriculture, forestry,
       fishing (AFOFI) (IPCC category 1A4a, 1A4b, 1A4c), and the transportation sector
       (IPCC category 1A3), excluding pipeline emissions (IPCC category 1A3ei),
-      not accounting for negative emissions from bioenergy with CCS (BECCS) in these sectors
+      not accounting for emissions removed from the atmosphere
```
znichollscr, line 192:
```suggestion
      not accounting for removals from the atmosphere
```

Alright I'll stop writing this now, but I think you get the point. You're not really removing "emissions", you're removing "carbon" so just writing "removals" might be simpler

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +278,24 @@
...
+      through incineration or decomposition (IPCC 1996 category 6; IPCC 2006 category 4), 
+      not including emissions from organic waste handling and decay
     unit: "{Level-2 Species}"
 
 - Emissions|{Level-2 Species}|Other:
-    description: Emissions of {Level-2 Species} from other sources
+    description: Emissions of {Level-2 Species} from other sources (IPCC 1996 category 7;
+      IPCC 2006 category 5B) including fossil fuel fires and any sources or sinks that
+      cannot be accommodated in other categories
+      notes: This will be treated as a flux from some fossil reservoir to the atmosphere in climate models.
```
znichollscr, line 290:
```suggestion
      notes: This will be treated as a flux from the fossil reservoir to the atmosphere in climate models.
```

Might be clearer this way

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +278,24 @@
...
-      e.g., direct air capture (DAC) or enhanced weathering
+- Emissions|{Level-3 Species}|Other Capture and Removal:
+    description: Capture and removal of atmospheric {Level-3 Species} using other net-negative 
+      technologies that are not directly linked 
+      to an emissions source, including direct air carbon capture and storage (DACCS) or enhanced 
+      weathering (EW), as well as durable wood products in building elements, mineral products, 
+      ocean fertilization, biomass sinking, ocean alkalinity enhancement (OAE), direct ocean capture
     unit: "{Level-3 Species}"
     note: This timeseries should be reported as negative values so that subcategories of
       emissions add up to net emissions `Emissions|{Level-3 Species}`.
```
znichollscr, line 301: Is it worth saying something in the notes like, "for a breakdown of the kinds of removal, see the 'Removal\*' tree"

jkikstra:
Ready for review. 

Note: there's now a bunch of new AFOLU variables.

@IAMconsortium/common-definitions-land (e.g. @tmkhsgw @flohump @gidden) how is this for you?

danielhuppmann:
I just realized that there is some overlap with #179 - both this PR and #179 add **Gross Emissions|CO2|AFOLU**, but the subdivision below AFOLU|Land emissions is not aligned between new variables for net vs. gross emissions... Is that a problem? Should we try to keep net and gross emissions variables aligned all the way down the hierarchy?

@IAMconsortium/common-definitions-land @tmkhsgw @flohump @gidden

definitions/variable/emissions/emissions.yaml
```diff
@@ -248,21 +341,25 @@
...
+      through incineration or decomposition (IPCC 1996 category 6; IPCC 2006 category 4), 
+      not including emissions from organic waste handling and decay
     unit: "{Level-2 Species}"
 
 - Emissions|{Level-2 Species}|Other:
-    description: Emissions of {Level-2 Species} from other sources
+    description: Emissions of {Level-2 Species} from other sources (IPCC 1996 category 7;
+      IPCC 2006 category 5B) including fossil fuel fires and any sources or sinks that
+      cannot be accommodated in other categories
+      notes: This will be treated as a flux from the fossil reservoir to the atmosphere in climate models.
```
danielhuppmann, line 353: This will fix the failing test.
```suggestion
    notes: This will be treated as a flux from the fossil reservoir to the atmosphere in climate models.
```

flohump:
> I just realized that there is some overlap with #179 - both this PR and #179 add **Gross Emissions|CO2|AFOLU**, but the subdivision below AFOLU|Land emissions is not aligned between new variables for net vs. gross emissions... Is that a problem? Should we try to keep net and gross emissions variables aligned all the way down the hierarchy?
> 
> @IAMconsortium/common-definitions-land @tmkhsgw @flohump @gidden

As mentioned by @danielhuppmann this PR has quite some overlap with #179, where I also made an attempt to improve the definition of the AFOLU emission variable. 
I'm happy to withdraw my PR if the most important additions are adopted in this PR.

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,27 +1,55 @@
...
+      negative emissions, for example from bioenergy with CCS (BECCS), direct air carbon capture
+      or agriculture, forestry and other land use (AFOLU)
+    notes: For net emissions (accounting for negative emissions), see "Emissions|CO2".
     unit: Mt CO2/yr
 
 - Emissions|{Level-2 Species}|AFOLU:
     description: Emissions of {Level-2 Species} from agriculture, forestry
-      and other land use (IPCC category 3)
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3), 
+      reporting the sum of sources and sinks including land-use methods that do not
```
flohump, line 34: Please add examples for sources like:
In particular, this includes agricultural emissions (mostly non-CO2) from livestock, rice cultivation and soil management as well as emission from land use and land-use change (mostly CO2) such as deforestation, conversion of non-forest natural land and drained peatlands.

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,27 +1,55 @@
...
+      or agriculture, forestry and other land use (AFOLU)
+    notes: For net emissions (accounting for negative emissions), see "Emissions|CO2".
     unit: Mt CO2/yr
 
 - Emissions|{Level-2 Species}|AFOLU:
     description: Emissions of {Level-2 Species} from agriculture, forestry
-      and other land use (IPCC category 3)
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3), 
+      reporting the sum of sources and sinks including land-use methods that do not 
+      use CCS such as soil carbon sequestration, biochar, forestry, peatland
```
flohump, line 35: Pleaes remove "peatland and wetland restoration, and biomass burial"
These have been also removed from Carbon Removal.

flohump, line 35: please add Re/Afforestation and agroforestry as examples for sinks

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,52 +71,113 @@
...
+    components:
+      - Emissions|{Level-3 Species}|AFOLU|Land|Forest Land
+      - Emissions|{Level-3 Species}|AFOLU|Land|Cropland
+      - Emissions|{Level-3 Species}|AFOLU|Land|Grassland
+      - Emissions|{Level-3 Species}|AFOLU|Land|Wetlands
+      - Emissions|{Level-3 Species}|AFOLU|Land|Settlements
+      - Emissions|{Level-3 Species}|AFOLU|Land|Other Land
+      - Emissions|{Level-3 Species}|AFOLU|Land|Fires
+- Emissions|{Level-3 Species}|AFOLU|Land|Forest Land:
+    description: Emissions and removals from lands with woody vegetation
```
flohump, line 89: Does this include CO2 emission from deforestation? If so, pleae add it to the desc.

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,52 +71,113 @@
...
+      (partially IPCC 1996 category 5A/5B/5D; IPCC 2006 category 3B1) 
+    note: It also includes systems with vegetation that currently fall below, 
+      but are expected to later exceed, the threshold values used by a country to
+      define the forest land category.
+    unit: "{Level-3 Species}"
+- Emissions|{Level-3 Species}|AFOLU|Land|Cropland:
+    description: Emissions and removals from arable and tillage land, rice
```
flohump, line 96: Does this also include GHG emissions from drained peatlands used as cropland?

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,52 +71,113 @@
...
+      fields, and agro-forestry systems where vegetation falls
+      below the thresholds used for the forest land category
+      (no clear IPCC 1996 category; IPCC 2006 category 3B2)
+    note: This does not include methane emissions from rice cultivation
+    unit: "{Level-3 Species}"
+- Emissions|{Level-3 Species}|AFOLU|Land|Grassland:
+    description: Emissions and removals from rangelands and pasture land
```
flohump, line 103: Does this also include GHG emissions from drained peatlands?

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,52 +71,113 @@
...
+      in the forest land category and are not expected to exceed
+      them, without human intervention (partially IPCC 1996 category 4D/4E/5A/5B/5C/5D; IPCC 2006 category 3B3).
+    note: The category also includes all grassland from wild lands to recreational 
+      areas as well as agricultural and silvi-pastural systems, subdivided
+      into managed and unmanaged, consistent with national
+      definitions.
+    unit: "{Level-3 Species}"
 - Emissions|{Level-3 Species}|AFOLU|Land|Wetlands:
-    description: Emissions of {Level-3 Species} from managed peatlands (drained and rewetted)
+    description: Emissions from land that is covered or saturated by water for
```
flohump, line 114: This would be only GHG emissions from rewetted peatlands / wetlands?
Because GHG emissions intact peatlands / wetlands are not antrophogenic.
Or should all GHG emissions from drained and rewetted peatands be reported in this category?

flohump (changes requested):
Thanks for the revised definitions of AFOLU variables.
I have a couple of suggestions for further refinement based on #179

definitions/variable/emissions/emissions.yaml
```diff
@@ -43,52 +75,118 @@
...
+    notes: It also includes systems with vegetation that currently fall below, 
+      but are expected to later exceed, the threshold values used by a country to
+      define the forest land category.
+    unit: "{Level-3 Species}"
+- Emissions|{Level-3 Species}|AFOLU|Land|Cropland:
+    description: Emissions and removals from arable and tillage land, rice
+      fields, and agro-forestry systems where vegetation falls
+      below the thresholds used for the forest land category
+      (no clear IPCC 1996 category; IPCC 2006 category 3B2)
+    notes: This does not include methane emissions from rice cultivation
```
flohump, line 109: Why is CH4 from rice cultivation excluded? It also contradicts with the desc: "rice fields"

jkikstra:
@flohump should there be _more_ AFOLU sectors?
E.g. liming (3C2), harvested wood products (3D1), and others?
Or should we map them ourselves, directly onto the variables we already have?
I'm following: https://www.ipcc-nggip.iges.or.jp/public/2019rf/pdf/1_Volume1/19R_V1_Ch08_Reporting_Guidance.pdf

For a quick (but non-authorative?) overview see slide 17 of: https://unfccc.int/sites/default/files/1_1996_to_2006_lesotho_march_2016_final_kt.pdf

flohump:
> @flohump should there be _more_ AFOLU sector variables? E.g. liming (3C2), harvested wood products (3D1), and others? Or should we map them ourselves, directly _onto the variables we already have_? I'm following: https://www.ipcc-nggip.iges.or.jp/public/2019rf/pdf/1_Volume1/19R_V1_Ch08_Reporting_Guidance.pdf
> 
> For a quick (but non-authorative?) overview see slide 17 of: https://unfccc.int/sites/default/files/1_1996_to_2006_lesotho_march_2016_final_kt.pdf

Harvested Wood Products (HWP) would be good to have (sum of sources and removals) for contrasting gross removals in `Carbon Removal|Long-Lived Materials`

jkikstra:
Comments discussion with @flohump 

- keeping burning variables separate makes sense
- Split of 3B categories is probably not doable for IAM teams (e.g. aggregate deforestation can be done) - at least not for MAgPIE
- Agriculture description may be updated a bit
- Harvested Wood Products will be added
- We will go with the following split for Land:

    components:
      - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-use Change # includes deforestation and conversion from all other natural land, CDR: ref/aff, agrofor, timber, ...
      - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-use Change|Wetlands # drained, rewetted, (not intact)
      - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-use Change|Other # ...everything not in Wetlands...
      - Emissions|{Level-3 Species}|AFOLU|Land|Fires
      - Emissions|{Level-3 Species}|AFOLU|Land|Harvested Wood Products # IPCC 3D1
      - Emissions|{Level-3 Species}|AFOLU|Land|Other # IPCC 3D2


This would make us both happy

flohump:
I just noticed that tier level are currently missing from the emission reporting @jkikstra 
I think it would be useful to have tier level based on this definition from @danielhuppmann 
Tier-1: Must-have
Tier-2: Important for sensible analysis
Tier-3: Nice-to-have

jkikstra:
> I just noticed that tier level are currently missing from the emission reporting @jkikstra I think it would be useful to have tier level based on this definition from @danielhuppmann Tier-1: Must-have Tier-2: Important for sensible analysis Tier-3: Nice-to-have

I'm OK if we go ahead here without that for now and add it in the days after we send out the template update

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,94 +1,186 @@
...
+      This is reporting the sum of sources and sinks including land-use methods that do not 
+      use CCS such as soil carbon sequestration, biochar, forestry, and re/afforestation and 
+      agroforestry. 
     unit: "{Level-2 Species}"
+    notes: Emissions reported in this category represent fluxes to/from the atmosphere from/to the land pool. 
+      For a breakdown of carbon removals, see the "Carbon Removal|*" tree
+    components:
+      - Emissions|{Level-3 Species}|AFOLU|Agriculture
+      - Emissions|{Level-3 Species}|AFOLU|Agricultural Waste Burning
+      - Emissions|{Level-3 Species}|AFOLU|Land
```
danielhuppmann, line 47:
```suggestion
      - Emissions|{Level-2 Species}|AFOLU|Agriculture
      - Emissions|{Level-2 Species}|AFOLU|Agricultural Waste Burning
      - Emissions|{Level-2 Species}|AFOLU|Land
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,94 +1,186 @@
...
+- Gross Emissions|CO2|AFOLU:
+  description: Gross emissions of carbon dioxide (CO2)  from agriculture, forestry
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3).
+      This is emissions sources only. All removals (for example, 
+      removals from biochar) are not included here. For net emissions, 
+      see "Emissions|*". Emissions reported in this category represent
```
danielhuppmann, line 53:
```suggestion
      see "Emissions|CO2|AFOLU". Emissions reported in this category represent 
```

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,94 +1,186 @@
...
+      of new cropland is reported under "Emissions|*|AFOLU|Land|Cropland".
+    components:
+      - Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change
+      - Emissions|{Level-3 Species}|AFOLU|Land|Fires
+      - Emissions|{Level-3 Species}|AFOLU|Land|Harvested Wood Products
+      - Emissions|{Level-3 Species}|AFOLU|Land|Other
+- Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change:
+    description: Emissions and removals from land use and land-use change,
+      including emissions from deforestation and conversion from all other natural land,
+      as well as removals from the atmosphere through agroforestry, re/afforestation
```
danielhuppmann, line 105:
```suggestion
      as well as removals from the atmosphere through agroforestry or re/afforestation
```

jkikstra:
Feedback from the IAMC SWG Data - Emissions group meeting, which will be discussed a bit more.

Major:
- Should we report Fires separate from AFOLU? Two reasons were brought up by @shinichirofujimoriKU :
    - Fires are not clearly just anthropogenic, but a mix with also some emissions from natural fires.
    - Fires are gridded separately (from agriculture)
Upon reflection: I think this is probably a good suggestion. 
Note on harmonization (for SCMs and ESMs): peat fires and deforestation & degradation are represented in the GCB, and are also the ones that should be included in AFOLU-like forcing (but not other fires!) following the guidance of [Van Marle et al. (2017)](https://gmd.copernicus.org/articles/10/3329/2017/gmd-10-3329-2017.pdf) "Climate models should not use the CO2 emissions (or nitrogen emissions if the nitrogen cycle is included in the model), as forcing because in general these emissions are not net emissions to the atmosphere, but a return pathway of previously sequestered carbon just as respiration is. The exceptions are CO2 emissions from deforestation and peat fires.".
- Can we add a signpost for national accounting vs scientific model reporting? (@gidden offered to come with a suggestion)

Minor: 
- Can it be clarified a bit more cleanly how Harvested Wood Products (HWP) should be reported? Maybe have a look at Chapter 12 of Volume 4 of the 2019 refinement of the reporting guidelines and report on a preferred suggested way?

gidden:
> Can we add a signpost for national accounting vs scientific model reporting? (@gidden offered to come with a suggestion)

Simple suggestion is to use [tags] like [model] or [nghgi]

jkikstra:
> nghgi

Do we need an extra variable?

jkikstra:
> I just noticed that tier level are currently missing from the emission reporting @jkikstra I think it would be useful to have tier level based on this definition from @danielhuppmann Tier-1: Must-have Tier-2: Important for sensible analysis Tier-3: Nice-to-have

P.S. in the presentation that the SWG Data leads were giving now, they put
Tier 1: Basic variables, keep number as small as possible
Tier 2: Standard reporting
Tier 3: Detailed reporting, enabling in-depth analysis

gidden:
> > nghgi
> 
> 
> 
> Do we need an extra variable?

The use case here is that teams would be required to report "LULUCF [model]" and can optionally report in addition "LULUCF [nghgi]"

jkikstra:
On nghgi:
Suggestion coming from discussion between @gidden @jayfuhrman and me:
1. add extra AFOLU NGHGI variable (and make clear it is NOT a component of totals)
2. add a note to the model/scientific AFOLU variable that this is model/scientific accounting, and NOT NGHGI-aligned

This way, we don't break post-processing scripts, but still add the extra reporting detail option.

jkikstra:
for fires: we try discussing between @shinichirofujimoriKU and @gidden and the CMIP7 fires emissions people for an extra discussion

jkikstra:
@IAMconsortium/common-definitions-emissions ping to check before we merge?

znichollscr:
> * Should we report Fires separate from AFOLU?

Another reason is that we can maybe harmonise better if we do this. There are biomass burning emissions from GFED, so if we report separately, we can think about whether we want to do something more sophisticated than just harmonising to GCP at the global level

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,94 +1,179 @@
...
     description: Emissions of {Level-2 Species} from agriculture, forestry
-      and other land use (IPCC category 3)
+      and other land use (IPCC 1996 category 4, 5; IPCC 2006 category 3).
+      In particular, this includes agricultural emissions (mostly non-CO2) 
+      from livestock, rice cultivation and soil management as well as 
+      emissions from land use and land-use change (mostly CO2) such as deforestation, 
+      conversion of non-forest natural land, and drained peatlands. 
+      This is reporting the sum of sources and sinks including land-use methods that do not 
+      use CCS such as soil carbon sequestration, biochar, forestry, and re/afforestation and 
+      agroforestry.
```
znichollscr, line 40:
```suggestion
      This reports the sum of sources and sinks including land-use methods that do not 
      use CCS such as soil carbon sequestration, biochar, forestry, and re/afforestation and 
      agroforestry. 
```

My immediate question here would be, where does stuff that does use CCS go? Not sure if that's worth clarifying or not (the notes already clarifies a bit, but not unambiguously).

jkikstra:
> > * Should we report Fires separate from AFOLU?
> 
> Another reason is that we can maybe harmonise better if we do this. There are biomass burning emissions from GFED, so if we report separately, we can think about whether we want to do something more sophisticated than just harmonising to GCP at the global level

I think in principle this can be done anywhere where it would be reported, but sure. 


Just for everyone @IAMconsortium/common-definitions-emissions (e.g. @strefler ) - since @danielhuppmann wants to send out the new variable template today, I would really like to merge this in before. 
Then we can update after. 
I think that the current state of the template (+[comment](https://github.com/IAMconsortium/common-definitions/pull/188#issuecomment-2456185885)) is better than not merging this in before. 

And so we'll make a new PR directly afterwards then, to continue discussions?

znichollscr (approved):
Last thoughts, looks super nice thanks!

definitions/variable/emissions/emissions.yaml
jkikstra: Do you have examples of CCS on AFOLU?

znichollscr: > Do you have examples of CCS on AFOLU where it would not be clear that it should go under this variable?

I meant more the other way around. If I have a variable with CCS, I would not know where to put it (I can go looking through all the other categories, but we can also just say, "Look at the energy tree").

> Or should we just delete "that do not use"

Would also be fine

jkikstra:
My suggestion for now:
we take what we have (with final polishing fixes), and leave the changes to Fires reporting for a new PR (see #200).

Thumbs up from people for that?

definitions/variable/emissions/emissions.yaml
jkikstra: OK I did the latter in c4d4cc91c2f29750c1a0e6aeed7d8dd375884480

definitions/variable/emissions/emissions.yaml
```diff
@@ -1,94 +1,191 @@
...
     unit: "{Level-3 Species}"
+- Emissions|{Level-3 Species}|AFOLU|Agricultural Waste Burning:
+    description: Emissions from agricultural waste burning
+      (IPCC 1996 category 4F; partially IPCC 2006 category 3C1)
+    unit: "{Level-3 Species}"
+    notes: Also sometimes known as 'biomass burning' or 'open agricultural burning'
 - Emissions|{Level-3 Species}|AFOLU|Land:
-    description: Emissions of {Level-3 Species} from forestry and other land use
+    description: Emissions of {Level-3 Species} from forestry and other land use and 
+      land use change. Removals in this variable include agroforestry, re/afforestation,
```
strefler, line 93: Looks to me like all the removal options are missing as subcategories. Not a showstopper though, can add them later, just need to ensure consistency with the carbon management variables

strefler:
great, thanks! Let's merge this and leave other updates for later PRs

definitions/variable/emissions/emissions.yaml
jkikstra: That's right, thanks - I'll note it down for future refinement.

danielhuppmann (approved):
Thank you @jkikstra 

Merging as is so that ScenarioMIP-SSP-submission can proceed, but this will require more discussion in several variables and a harmonization of descriptions.

PR #190, opened 2024-10-28
Add variables for energy prices

danielhuppmann:
This PR adds investment variables suggested by @gunnar-pik for energy prices at the relevant levels.

FYI @IAMconsortium/common-definitions-energy

### DIFF
--- a/definitions/variable/energy/energy-prices.yaml
+++ b/definitions/variable/energy/energy-prices.yaml
@@ -1,0 +1,27 @@
+- Price|Primary Energy|Biomass:
+    description: Price for biomass
+    unit: USD_2010/GJ
+    weight: Primary Energy|Biomass
+- Price|Primary Energy|{Primary Fossil Fuel}:
+    description: Price for {Primary Fossil Fuel} at global or regional spot markets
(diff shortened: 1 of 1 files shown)


masahirosugiyama:
@danielhuppmann Thank you for this. Just wondering, why do we continue to use USD_2010? I mean, why 2010, not 2015 or 2020? Perhaps we can specify it with a tag file so that we can change it later?

danielhuppmann:
Thanks, but not sure if this would make much difference in practice... 

And you can easily convert for analysis. With pyam, it's as simple as
```
df.convert_unit("USD_2010", "USD_2020")
```
using conversion factors from the [iam-units](https://github.com/IAMconsortium/units/blob/e22eab633e1aa65ac89ab110644396fab10aa77f/iam_units/data/definitions.txt#L63) package.

If you still feel strongly about this, please start a new issue.

pkyle:
At least 2010 is modern! GCAM still uses 1975 dollars because the model was first made in the early 1980s 😆

masahirosugiyama:
Wow, didn’t know this about GCAM… 
OK, now understood.

PR #192, opened 2024-10-29
Update sdg.yaml

VassilisDaioglou:
Added Population| 

- Sanitation Access
- Drinking Water Access

### DIFF
--- a/definitions/variable/sdg-indicators/sdg.yaml
+++ b/definitions/variable/sdg-indicators/sdg.yaml
@@ -37,7 +37,6 @@
 - Population|Drinking Water Access:
     description: Population with access to improved drinking  water sources 
     unit: million
-    shape: Population|Drinking Water Access
 - Expenditure|Households|Food [Share]:
     description: Share of food expenditure in total income
     unit: '%'
--- a/definitions/variable/sdg-indicators/sdg.yaml
+++ b/definitions/variable/sdg-indicators/sdg.yaml
@@ -34,6 +34,9 @@
 - Population|Sanitation Access:
     description: Population with access to improved sanitation
     unit: million
+    navigate: Population|Sanitation Acces
+    engage: Population|Sanitation Acces
+    shape: Population|Sanitation Acces
 - Population|Drinking Water Access:
     description: Population with access to improved drinking  water sources 
     unit: million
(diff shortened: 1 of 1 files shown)


definitions/variable/sdg-indicators/sdg.yaml
```diff
@@ -31,6 +31,14 @@
       gas)
     unit: million
     navigate: Population|Clean cooking Access
+- Population|Sanitation Access:
+    description: Population with access to improved sanitation
+    unit: million
+    shape: Population|Sanitation Access
```
danielhuppmann, line 37: Reference to "shape" (or other projects) can be omitted if the variable name did not change. This is only intended for automated mapping across projects.
```suggestion
```

definitions/variable/sdg-indicators/sdg.yaml
```diff
@@ -31,6 +31,14 @@
...
+- Population|Drinking Water Access:
+    description: Population with access to improved drinking  water sources 
+    unit: million
+    shape: Population|Drinking Water Access
```
danielhuppmann, line 41: Can be removed as above.
```suggestion
```

danielhuppmann (commented):
Two comments inline. I'm wondering I we shouldn't move these variables "access to ..." to an own (sub)-category.

VassilisDaioglou:
RE moving to a different sub-category could be something to be considered. eIn the future a number of "Access"variables will beome important. 

They all fall under "Population|Access|[...]", so it could also fall under the "Population" sub-category with a "tag_access".

definitions/variable/sdg-indicators/sdg.yaml
```diff
@@ -31,6 +31,12 @@
       gas)
     unit: million
     navigate: Population|Clean cooking Access
+- Population|Sanitation Access:
+    description: Population with access to improved sanitation
+    unit: million
```
danielhuppmann, line 36: Sorry, just realized that this variable actually did exist, just misspelled..
```suggestion
    unit: million
    navigate: Population|Sanitation Acces
    engage: Population|Sanitation Acces
```

danielhuppmann:
Well, I thought maybe moving this to a new top-level category like "Decent Living Indicator" - we already have too many items categorized under "Population". But not sure if that's a good move right now...

danielhuppmann:
@VassilisDaioglou, can you commit my proposed change to add the mapping to the mis-spelt variables in earlier projects, then I'll merge.

VassilisDaioglou:
Done - sorry for the delay.

danielhuppmann (approved):
Thanks, looks good!

PR #180, opened 2024-10-22
LU afolu variables

flohump:
This revision of reporting variables is part of a larger set of changes related to AFOLU reporting variables that have been discussed and agreed with representative of all modeling teams in @IAMconsortium/common-definitions-land 
The revision is based on https://1drv.ms/x/s!AlN2Seu8OCz8hLQ8Wh_ysgHyDSg4QA
For consistency with naming conventions some variables have been named and defined differently.
The purpose of tier, domain and subdomain is to allow for automated checks of reported variables in specific domains.
Due to the use of tag yaml files some tier levels could not be set as discussed. 
@danielhuppmann Is it possible to use tier levesl in tag yaml files?

PR #195, opened 2024-10-30
new folder for "environment" variables

flohump:
As discussed [here](https://github.com/IAMconsortium/common-definitions/pull/180#discussion_r1819594078), this PR renames the `water` folder to `environment` and moves nutrients and biodiversity variables from `sdg-indicators` to the new `environment` folder.
Moreover, `Nitrogen|Cropland Budget|Input` is renamed and simplified to `Fertilizer Use|Nitrogen`
Biodiversity indicators are copy-paste from `sdg-indicators`.


jonathandoelman (approved):
Yes, great improvement in my opinion! I dont see any technical issues or errors either.

tmkhsgw:
Great! 
Some comments:
I guess we need a tag for Electricity Source which is used in water.yaml.

- Fertilizer Use|Nitrogen:
- Fertilizer Use|Nitrogen|Synthetic:
- Fertilizer Use|Nitrogen|Organic:
Unit should be Tg N/yr, not Tg P/yr.

flohump:
> Great! Some comments: I guess we need a tag for Electricity Source which is used in water.yaml.
> 
> * Fertilizer Use|Nitrogen:
> * Fertilizer Use|Nitrogen|Synthetic:
> * Fertilizer Use|Nitrogen|Organic:
>   Unit should be Tg N/yr, not Tg P/yr.

Thanks for spotting my copy-paste mistake. Fixed now. 
Tags also work if defined in other folders. So, no need to move them to the `environment` folder.

VassilisDaioglou (approved):

definitions/variable/environment/biodiversity.yaml
```diff
@@ -0,0 +1,64 @@
...
+    weight: Land Cover
+    navigate: Terrestrial Biodiversity|BII
+    tier: 2
+- Terrestrial Biodiversity|Biodiversity Intactness Index|Cropland Landscapes:
+    description: Terrestrial biodiversity in landscapes containing cropland measured
+      with Biodiversity Intactness Index (BII)
+    unit: '%'
+    sdg: 15
+    skip-region-aggregation: true
+    weight: Land Cover|Cropland
```
danielhuppmann, line 35: This is copy-pasting a mistake already in the sag.yaml file - specifying a weight but also stating to skip the region-aggregation is a bit of a contradiction (in nomenclature, the **skip-region-aggregation** specifier wins). 

Are you sure you want to skip the region aggregation if you know the weight?

definitions/variable/environment/biodiversity.yaml
```diff
@@ -0,0 +1,64 @@
+- Terrestrial Biodiversity|Mean Species Abundance:
+    description: Terrestrial biodiversity measured in Mean Species Abundance (MSA)
+    unit: '%'
+    sdg: 15
+    skip-region-aggregation: true
+    weight: Land Cover
```
danielhuppmann, line 6: As below, this is a contraction...

danielhuppmann (commented):
I'm loving it how everyone is reviewing within a few hours!

Moving variables over from one file to the other allowed me to spot a pre-existing issue, see the comments inline.

definitions/variable/environment/biodiversity.yaml
flohump: Thanks for spotting. I changed to weight to "Land Cover" for all biodiversity variables. Otherwise, I see the risk that results are not comparable.

definitions/variable/environment/biodiversity.yaml
flohump: revised

danielhuppmann (approved):
Thank you!

PR #198, opened 2024-11-01
Add ScenarioMIP-CMIP7 emissions variables

danielhuppmann:
This PR adds variables that were "manually" added to the ScenarioMIP reporting template, see https://github.com/iiasa/ssp-submission-workflow/blob/main/definitions/variable/variables.yaml.


jkikstra (approved):
I don't have any objections to adding these variables, but just to note that I don't think that these variables have been used for ScenarioMIP last time and also are unlikely to be used for ScenarioMIP this round. 
Still, fine to merge in as the extra level of detail is probably useful to some users and gives a bit more flexibility later.

PR #199, opened 2024-11-02
Add sectors for AFOLU under Gross Emissions 

jkikstra:
Per suggestion of @tmkhsgw, see https://github.com/IAMconsortium/common-definitions/pull/188#issuecomment-2445868743.

Moved into this PR following suggestion of @danielhuppmann; https://github.com/IAMconsortium/common-definitions/pull/188#discussion_r1826515740.


jkikstra:
This probably should be closed, as #179 should include these updates

PR #200, opened 2024-11-05
Fires Emissions: changing hierarchy

jkikstra:
Note: do not merge before #188 .

This PR is a first attempt to try to accommodate the suggestion of @shinichirofujimoriKU, as described here: https://github.com/IAMconsortium/common-definitions/pull/188#issuecomment-2456156210


jkikstra:
Two thoughts:
1. (major) I'm not sure I'm convinced about placing all GFED-based burning outside of AFOLU, mainly because to me AWB clearly seems related to Agriculture
2. (minor) I did not touch fossil fuel fires, to keep it aligned with EDGAR&CEDS reporting under "Other", and because no IAM models this AFAIK. (but of course, in principle, this is also Fire/Burning)

znichollscr:
> 1. (major) I'm not sure I'm convinced about placing all GFED-based burning outside of AFOLU, mainly because to me AWB clearly seems related to Agriculture

I agree with this, although I think it's a general problem of the tree approach. We have a category that could have AFOLU or burning as a higher-level. In a tree, there's no easy way to have two parent nodes, so we're just going to have to live with whatever and maybe provide a mapping between the two if we find ourselves doing this conversion a lot.

shinichirofujimoriKU:
I am not sure it has been resolved already, but I think AWB is purely anthropogenic which would be OK to be reported under AFOLU. My main point is better to keep the variables which can include non-anthropogenic emissions separate. 
At the SWG meeting, I mentioned the downscaling process as one of the reasons to separate fire related emissions, but probably distinguishing treatment of anthropogenic (AWB) or not (other fire) would be useful. 
downscaling and MAGICC run can be done regardless of the decision how subcategory is defined.

jkikstra:
> I am not sure it has been resolved already, but I think AWB is purely anthropogenic which would be OK to be reported under AFOLU. My main point is better to keep the variables which can include non-anthropogenic emissions separate. 

So, let me try to hone that suggestion, and check your response to the split now in https://github.com/IAMconsortium/common-definitions/pull/200/commits/bf9270b766ce1df214e41ebc9f8e829403930649

> At the SWG meeting, ...I mentioned the downscaling process as one of the reasons to separate fire related emissions, but probably distinguishing treatment of anthropogenic (AWB) or not (other fire) would be useful. downscaling and MAGICC run can be done regardless of the decision how subcategory is defined.

Agreed. Same holds for (not only for downscaling and MAGICC, but also for) spatial gridding. 
I would think we do separate grids for Peat Fires, Forest Fires, and Grassland Fires either way (no matter where we place it).

PR #99, opened 2024-05-24
Add initial techno-economic variables

rongqizhu:
Dear common-definitions-techno-economic group,]

I am excited to be part of techno-economic team in common-definitions project. This PR is an initial implementation about techno-economic variables.

I have solved most of the problems from [https://github.com/IAMconsortium/common-definitions/pull/71](url), like renaming "OM Cost" to "Operation and Maintenance Cost”,  renaming "US$2010" to "USD_2010" .

Next, we will make further improvement about the stucture of variables "OM Cost|Fixed|...", discussing with the team!

PR #201, opened 2024-11-05
Add variables for capital cost, efficiency and technical lifetime

danielhuppmann:
This PR carries over the list of techno-economic variables from #99 by @rongqizhu, which was also requested by @gunnar-pik separately by email.

The proposed variable for variable vs. fixed operation and maintenance cost is not implemented here, this will be discussed in a follow-up issue and PR.


danielhuppmann:
fyi @IAMconsortium/common-definitions-energy

PR #208, opened 2024-11-12
Add variables for carbon capture by source

danielhuppmann:
This PR implements changes discussed by email among the emissions-group (see in particular the email by @strefler on October 10).

The changes in particular:
- add variables for carbon capture by source (fossil, biomass, synthetic, processes)
- add tiers and bounds for existing carbon-capture variables
- add the list of components of the top-level "Carbon Capture" variable to clarify that "Carbon Capture|Fossil" and "Carbon Capture|Geological Storage" should not be added up

[EDIT by @gidden]: The October 10 email includes the following attachment [MissingVariables-CDR-CCS.xlsx](https://github.com/user-attachments/files/17747560/MissingVariables-CDR-CCS.xlsx)

[EDIT by @gidden]:

We propose 3 main variable elements:

1. the "reservoir tree" denoting the provenance of carbon in each storage reservoir (`Carbon Capture|<Storage Reservoir>|<Carbon Source>`)
2. the "full tree" denoting in which sectors carbon is captured, the type of carbon captured, and the final resting place of that carbon (`Carbon Capture|<Sector>|<Carbon source>|<Storage Reservoir>` )
3.  the "source tree" denoting at a high level source totals (`Carbon Capture|<Carbon source>`)

Reservoirs include:
- Geologic storage
- Other storage
- Leakage and other Losses
- Utilization

Sources include:
- Biomass (or biogenic)
- Fossil
- Synthetic Fuel
- Direct Air Capture
- Direct Ocean Capture
- Industrial Processes
- Other Sources

Sectors include all relevant sectors from Energy, Industry, etc.

Key overlaps with Carbon Removal:

- Carbon Removal|Geologic Storage will be strictly less or equal to Carbon Capture|Geologic Storage (as it does not include fossil sources)
- Carbon Removal|Geological Storage|Bioenergy, Carbon Removal|Geological Storage|Direct Air Capture, and Carbon Removal|Geological Storage|Synthetic Fuels will have 1-to-1 analogues with the Carbon Capture reservoir tree
- Carbon Removal|Long-Lived Materials will equal the sum of non-fossil contributions to Carbon Capture|Utilization

Key other considerations
- Carbon capture only refers to energy and industrial processes, not, e.g., Forestry
- We will not include in the Capture tree silicate, weathering, and ocean-based methods

### DIFF
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -27,6 +27,18 @@
     tier: 1
     lower_bound: 0
     notes: Values should be reported as positive numbers.
+- Carbon Capture|Direct Air Capture:
+    description: Captured carbon dioxide (CO2) from direct air capture
+    unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+- Carbon Capture|Other:
+    description: Captured carbon dioxide (CO2) from other sources
+    unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
 
 - Carbon Capture|Geological Storage:
     description: Captured carbon dioxide (CO2) that is injected into subsurface geologic
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -33,6 +33,12 @@
     tier: 1
     lower_bound: 0
     notes: Values should be reported as positive numbers.
+- Carbon Capture|Direct Ocean Capture:
+    description: Captured carbon dioxide (CO2) from direct ocean capture
+    unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
 - Carbon Capture|Other:
     description: Captured carbon dioxide (CO2) from other sources
     unit: Mt CO2/yr
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -1,10 +1,7 @@
 - Carbon Capture:
-    description: Total amount of captured carbon dioxide (CO2) from energy use in supply
-      and demand sectors (IPCC category 1A, 1B) and from industrial processes
-      (IPCC categories 2A, B, C, E), including emissions from international bunkers,
-      net of negative emissions from bioenergy with CCS (BECCS),
-      direct air carbon capture and storage (DACCS), and other CO2 removal exclusive of
-      land-use change in these sectors
+    description: Total amount of carbon dioxide (CO2) captured either from the atmosphere,
+      directly from oceans, or from point sources as energy use in supply and demand
+      sectors (IPCC category 1A, 1B) or industrial processes (IPCC categories 2A, B, C, E)
     unit: Mt CO2/yr
     tier: 1
     lower_bound: 0
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -1,5 +1,3 @@
-# Carbon Capture is grouped into two levels of aggregation based on the fate of the captured CO2
-# storage versus utilization
 - Carbon Capture:
     description: Total amount of captured carbon dioxide (CO2) from energy use in supply
       and demand sectors (IPCC category 1A, 1B) and from industrial processes
@@ -8,6 +6,14 @@
       direct air carbon capture and storage (DACCS), and other CO2 removal exclusive of
       land-use change in these sectors
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    components:
+      - Carbon Capture|Geological Storage
+      - Carbon Capture|Other Storage
+      - Carbon Capture|Leakage
+      - Carbon Capture|Utilization
+
 - Carbon Capture|{Carbon Capture Source}:
     description: Captured carbon dioxide (CO2) from {Carbon Capture Source}
     unit: Mt CO2/yr
@@ -26,21 +32,31 @@
     description: Captured carbon dioxide (CO2) that is injected into subsurface geologic
       storage reservoirs (e.g., in situ mineralization).
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
     notes: Captured CO2 that is reacted at the earth's surface (e.g., ex situ mineralization)
       should be reported under Carbon Capture|Other Storage
 - Carbon Capture|Other Storage:
     description: Other types of storage for captured carbon dioxide (CO2)
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
 - Carbon Capture|Leakage:
     description: Captured carbon dioxide (CO2) that is vented, leaked, or re-emitted
       during transport or after injection into storage
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
     notes: Leakage should be reported as positive values. Leakage of CO2 stored in
       previous years should be reported here.
 - Carbon Capture|Utilization:
     description: Captured carbon dioxide (CO2) that is used in the manufacture of fuels
       or materials, or to enhance the productivity of oil and/or natural gas reservoirs
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
 - Carbon Capture|Utilization|{Carbon Utilization Type}:
     description: Captured carbon dioxide (CO2) that is used to produce {Carbon Utilization Type}
     unit: Mt CO2/yr
+    tier: 2
+    lower_bound: 0
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -8,6 +8,20 @@
       direct air carbon capture and storage (DACCS), and other CO2 removal exclusive of
       land-use change in these sectors
     unit: Mt CO2/yr
+- Carbon Capture|{Carbon Capture Source}:
+    description: Captured carbon dioxide (CO2) from {Carbon Capture Source}
+    unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+- Carbon Capture|Industrial Processes:
+    description: Captured carbon dioxide (CO2) from industrial processes
+      (e.g., cement production), not including energy use
+    unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
+    notes: Values should be reported as positive numbers.
+
 - Carbon Capture|Geological Storage:
     description: Captured carbon dioxide (CO2) that is injected into subsurface geologic
       storage reservoirs (e.g., in situ mineralization).
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -58,7 +58,7 @@
     lower_bound: 0
 - Carbon Capture|Leakage:
     description: Captured carbon dioxide (CO2) that is vented, leaked, or re-emitted
-      during transport or after injection into storage
+      during transport
     unit: Mt CO2/yr
     tier: 1
     lower_bound: 0
--- a/definitions/variable/emissions/tag_carbon-capture-source.yaml
+++ b/definitions/variable/emissions/tag_carbon-capture-source.yaml
@@ -5,5 +5,17 @@
     - Fossil:
         description: fossil fuels including coal, natural gas and conventional/unconventional oil
     - Synthetic Fuels:
-        description: synthetic fuels produced from carbon-capture-and-utilization (CCU),
-          not derived from fossil fuels
+        description: synthetic fuels produced from carbon captured from fossil fuels and
+          directly or indirectly from the atmosphere,
+          e.g. using carbon-capture-and-utilization (CCU) from biomass or fossil fuels
+          or Direct Air Capture (DAC)
+    - Synthetic Fuels|Carbon-Neutral:
+        description: synthetic fuels produced from carbon captured directly or indirectly
+          from the atmosphere, e.g. using carbon-capture-and-utilization (CCU) from biomass
+          or Direct Air Capture (DAC), not derived from fossil fuels
+    - Synthetic Fuels|Fossil-Based:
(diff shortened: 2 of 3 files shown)


danielhuppmann:
Please keep the discussion in this PR focused on the items listed above. Open a new conversation about additional variables or subcategories that need to be added.

gidden:
Thanks @danielhuppmann. I tried to generate the template based on this PR using `nomenclature export-definitions` and got the attached excel sheet [pr.xlsx](https://github.com/user-attachments/files/17747595/pr.xlsx)

I guess I have done something wrong because the Carbon Capture tree looks as follows
image

danielhuppmann:
Thanks @gidden, no that's right - the carbon-capture sub-categories by destination (geological storage, synthetic fuels, leakage, utilization) already exist. This PR adds the sub-categories by source (biomass, fossil, industrial processes). This is based on an email sent by @strefler a month ago...

gidden:
Hm, then I don't understand the `ssp-submission_variable_template_20241106.xlsx`, which includes many more variables:

image

gidden:
Got it, thanks!

definitions/variable/emissions/carbon-management.yaml
```diff
@@ -1,5 +1,3 @@
-# Carbon Capture is grouped into two levels of aggregation based on the fate of the captured CO2
-# storage versus utilization
 - Carbon Capture:
     description: Total amount of captured carbon dioxide (CO2) from energy use in supply
```
strefler, line 2: Suggestion: Total amount of carbon dioxide (CO2) captured either from the atmosphere, directly from oceans, or from point sources as energy use in supply and demand sectors (IPCC category 1A, 1B) or industrial processes

definitions/variable/emissions/carbon-management.yaml
```diff
@@ -8,25 +6,57 @@
...
     description: Other types of storage for captured carbon dioxide (CO2)
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
 - Carbon Capture|Leakage:
     description: Captured carbon dioxide (CO2) that is vented, leaked, or re-emitted
       during transport or after injection into storage
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
```
gidden, line 49: Leakage needs to be negative, right?

definitions/variable/emissions/carbon-management.yaml
```diff
@@ -8,25 +6,57 @@
...
     notes: Captured CO2 that is reacted at the earth's surface (e.g., ex situ mineralization)
       should be reported under Carbon Capture|Other Storage
 - Carbon Capture|Other Storage:
     description: Other types of storage for captured carbon dioxide (CO2)
     unit: Mt CO2/yr
+    tier: 1
+    lower_bound: 0
 - Carbon Capture|Leakage:
```
gidden, line 44: I would suggest to call this `Leakage or other Losses` and for the variable to be negative given the directionality of other capture variables

definitions/variable/emissions/carbon-management.yaml
```diff
@@ -8,25 +6,57 @@
...
     description: Captured carbon dioxide (CO2) that is vented, leaked, or re-emitted
```
strefler, line 45: "or after injection into storage" should be removed here. Otherwise the sums don't add up. Need to think about where to add this, potentially as a separate variable? But this is not urgent.

definitions/variable/emissions/tag_carbon-capture-source.yaml
```diff
@@ -0,0 +1,9 @@
+- Carbon Capture Source:
+    - Biomass:
+        description: purpose-grown bioenergy crops, biogas, crop and forestry residue,
+          municipal solid waste bioenergy and traditional biomass
+    - Fossil:
+        description: fossil fuels including coal, natural gas and conventional/unconventional oil
+    - Synthetic Fuels:
+        description: synthetic fuels produced from carbon-capture-and-utilization (CCU),
+          not derived from fossil fuels
```
strefler, line 9: We are missing Industrial Processes, Direct Air Capture and Other.

strefler (changes requested):
Addition of the three sources Direct Air Capture, Industrial Processes and Other seems essential to me

gidden:
> Addition of the three sources Direct Air Capture, Industrial Processes and Other seems essential to me

Thanks @strefler - I edited the top-level description with a proposal for how this PR should look in the end. What do you think?

definitions/variable/emissions/carbon-management.yaml
strefler: no, should also be positive, otherwise the sums don't add up

kriegler100:
Thank you, Matt, for the proposal on the PR's variable list. I also discussed the topic with Jessica today, so I hope my comments will be aligned with hers.
 
I agree with the general variable list, my comments are one larger issue and a set of minor items (I am omitting the "bioenergy" naming issue as this is discussed in issue 49):

**Larger issue:**  The Carbon Capture|Industry variables are confusing. Industry is certainly a sector to zoom in on wrt carbon capture, but for the sum over sources, the energy part is covered by Fossil+Biomass+Synthetic Fuels and we are left with Industrial processes as the only additional source of carbon capture from industry. So all variables Carbon Capture|Industry|Industrial Processes (Lines 75-80) should have the "|Industry" in the name removed, and directly sum to CC|Industrial Processes. Suggest to add CC|Industrial Processes|Other Sector to make sure the summation is complete.

Should we remove all other variables CC|Industry (Lines 30-33 and 51-74)? 

Note that 30-32 are captured by CC|Biomass+Fossil|Energy|Demand|Industry so those definitely can be removed. We should add CC|Synthetic Fuels|Energy|Demand|Industry to capture variable 33 and get the sum in Line 30 right. Hence we do not need 30-33, but can calculate them from the above if needed. 

On Lines 51-74, I do not see that further sectoral detail on carbon capture from industry energy use would be super useful. Hence we might indeed remove these variables. But if we keep them, let us use consistent nomenclature, i.e. CC|Biomass+Fossil+ Synthetic Fuel|Energy|Demand|Industry|<Sector>.  Then everything sums nicely. 

**Minor items:** 

M1.  Line 29 (xls spreadsheet): Carbon capture|Carbon-Neutral Fuels. There are two types of synthetic fuels, carbon-neutral and from fossil sources. If we want to have summation by source, I think we should interprete CC|Biomass, CC|Fossil, CC|DAC as everything not going through the synthetic fuel route, and then have CC|Synthetic Fuels and the split CC|Synthetic Fuels|Fossil and CC|Synthetic Fuels|Carbon-Neutral (the latter includes e-fuels using biogenic CO2 or DAC) as additional items in the source list. Then the summation works.    

M2. Line 50: Should be called: Carbon Capture|Other Sources 

M3. We probably need to add CC|Geological Storage|Synthetic Fuels and the further split into Carbon-Neutral and Fossil Synthetic Fuels see above to make the Geological Storage variables sum properly. This would replace CC|Geological Storage|Electricity in Line 84 (to be removed). 

M4. It would be good to add the sub-categories of Industrial processes to CC|Geological Storage|Industrial Processes (Line 85). E.g. it is interesting to get an idea about Cement CCS (cement ~ Non-metallic minerals) vs. captured CO2 from cement plants that is utilized. 

**Some broader points:**

B1. We need to define the boundary between short-lived and long-lived material. I suggest that everything stored in materials and re-released within 5 years (one IAM time step) is short-lived and everything which is stored for a decade or longer is long-lived. 

For completeness, we would need an emissions variable for the CO2 re-released from long-lived materials, i.e. Emissions|CO2|Materials|Long-lived. 

B2. Jessica already commented that Leakage should not include leakage from geological storage reservoirs, as this is not adding to the carbon captured per time step. I suggest that leakage from reservoirs is rather added as emissions variable: Emissions|CO2|Geological Storage|Leakage. 

B3, To fully establish the link to emissions, we would further need to add Emissions|CO2|Carbon Capture|Leakage (= Carbon Capture|Leakage) and Emissions|CO2|Materials|Short-lived (=Carbon Capture|Utilization|Materials|Short-Lived)

B4. The GB reports carbon removal from the cement carbonation sink. We also should include this in the reporting, this is 0.8 GtCO2 globally per year. Should we add Carbon removal|Cement carbonation to the carbon removal variables? The question is if then want two sums for carbon removal, one excluding and one including the cement carbonation sink.

danielhuppmann:
@strefler, added a variable for direct air capture and other sources. A variable for capture from industrial processes was already added in the initial commit of this PR.

Good to be merged?

definitions/variable/emissions/carbon-management.yaml
danielhuppmann: Done.

definitions/variable/emissions/carbon-management.yaml
danielhuppmann: Not implemented following the other comment by @strefler

definitions/variable/emissions/carbon-management.yaml
danielhuppmann: Done.

definitions/variable/emissions/tag_carbon-capture-source.yaml
danielhuppmann: Added directly as variables in `carbon-management.yaml`

danielhuppmann:
> 1. I think direct ocean capture is still missing, right?

Added.

> 2. I'm a bit confused now, do I understand correctly that some carbon capture sources are in tag_geological-storage, while others are directly in the emissions file? If so, I guess it would make sense to move them all to tag_geological-storage, no?

There are currently two tag-files that partially overlap
- `tag_carbon-capture-source.yaml` which includes a "Fossil" item and is used in "Carbon Capture|..."
- `tag_geological-storage.yaml` which is used in "Carbon Removal|Geological Storage|..." and does not include fossil fuels

We are working on a feature to use "master" tag-lists with filtering, but for now, keeping two distinct tag lists is necessary.

However, I renamed "Carbon Removal|Geological Storage|Bioenergy" to "Carbon Removal|Geological Storage|Biomass" as suggested by @gidden based on the comment @kriegler100.

> 3. I think it would make sense to take up the comments on the descriptions already in this PR, or would you rather move that to a separate one? Sorry, I missed that in my last comment

Done. Please point out if I overlooked a comment.

definitions/variable/emissions/tag_carbon-capture-source.yaml
kriegler100: > Added directly as variables in `carbon-management.yaml`

Definition of synthetic fuels "synthetic fuels produced from carbon-capture-and-utilization (CCU), not derived from fossil fuels" does not work in my view. Synthetic fuels can be produced with CO2 from fossils as well as with CO2 from cement plants.  

Carbon capture|Synthetic Fuel  "Carbon captured in synthetic fuels"
Carbon capture|Synthetic Fuel|Carbon-neutral  "Carbon captured in synthetic fuels produced with carbon captured directly or indirectly (e.g. biogenic carbon) from the air "
Carbon capture|Synthetic Fuel|Fossil   "carbon captured in synthetic fuels produced with carbon captured in fossil fuel combustion processes"
Carbon capture|Synthetic Fuel|Industrial Processes   "synthetic fuels produced with carbon captured in industrial processes (e.g. cement production; excluding fossil fuel combustion)"

strefler: Agree with Elmar here. We have the choice of either having synthetic fuels as a separate category, or trying to allocate the synthetic fuels to the other sources, depending on the source of the carbon that was used. The latter can get very messy, especially if a model considers a circular economy, where the carbon is captured and used several times. That would lead to an infinite circle of use and re-use, and the only way I see for a clean accounting is to consider synthetic fuels as carbon-neutral and as their own category. Subcategories of synthetic fuels can in principle be done, but I would suggest to omit them as one of the subcategories would have to be synthetic fuels again...

kriegler100: Good point, Jessica. Since Synthetic Fuels can be both source as well as destination, they would show up both in Carbon Capture|<source> as well as Carbon Capture|Utilization|<type of use = synthetic fuel> and Carbon Capture|Geological Formation|<type of captured carbon = synthetic fuel>. 

However, I do not think we can restrict all of this to "carbon neutral". The problem of re-cycling is not solved by this and we are missing a significant part of synfuels that are not carbon neutral. Hence I suggest to stick with the subcategorization but be clearer in the definitions.

 "Carbon captured in synthetic fuels produced with carbon captured directly or indirectly (e.g. biogenic carbon or carbon from carbon-neutral synfuels) from the air"
Carbon capture|Synthetic Fuel|Fossil "carbon captured in synthetic fuels produced with carbon captured in fossil fuel combustion processes or from fossil-based synfuels"
Carbon capture|Synthetic Fuel|Industrial Processes    "carbon captured in synthetic fuels originally produced with carbon captured in industrial processes (e.g. carbon captured from cement production or from synfuels derived with carbon from industrial processes). Fossil-fuel based synthetic fuels are excluding here."

Of course, if synthetic fuels from different origins are mixed at some point, and then the carbon is captured from this mix, the separation becomes harder (would need to resort to fractions of the mix). But the same is true if we want to restrict just to "carbon neutral synfuels".

danielhuppmann:
Given this discussion https://github.com/IAMconsortium/common-definitions/pull/208#discussion_r1846940907 by @kriegler100 and @strefler, I separated synthetic fuels into atmospheric vs. fossil origin.

From my understanding, the aggregate of synfuel-capture (how much carbon is captured from fossil-based and carbon-neutral synthetic fuels) is of limited relevance for analysis, so I added them as two parallel variables, not sub-categories. (There is also the Secondary-Energy tree that has distinction between syngas vs. synthetic liquids with CCS).

definitions/variable/emissions/tag_carbon-capture-source.yaml
strefler: I was only suggesting to treat the synfuels as carbon neutral, and report the emissions where the CO2 is captured. But that is another discussion. We will most likely not be able to report the subcategories in REMIND, but there is also no harm in having the variables here. So agree with Elmars suggestion. After that we should be good to merge!

danielhuppmann:
Just realized that I hadn't pushed the commit, so now changed in light of @strefler's latest comment(https://github.com/IAMconsortium/common-definitions/pull/208#discussion_r1847934832 to have a synthetic-fuels item and have three sub-categories for the source of the carbon. Is that what you intended?

strefler (approved):

PR #209, opened 2024-11-12
Add variables for combustion and fugitive emissions of fuel supply

danielhuppmann:
This PR adds variables to distinguish between fugitive and combustion emissions at the energy-supply-by-fuel resolution, based on a discussion between @volker-krey and Steve Smith related to CEDS and CMIP6 harmonization.

Note that "Emissions|{Level-2 Species}|Energy|Supply|Combustion" and "Emissions|{Level-2 Species}|Energy|Supply|Fugitive" are already included in the template.

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -253,12 +253,12 @@
 - Emissions|{Level-2 Species}|Energy|Supply|{Secondary Fuel Level 1}|Combustion:
     description: Emissions of {Level-2 Species} from fuel combustion during production
       of {Secondary Fuel Level 1}
-    unit: "{Level-1 Species}"
+    unit: "{Level-2 Species}"
     tier: 2
 - Emissions|{Level-2 Species}|Energy|Supply|{Secondary Fuel Level 1}|Fugitive:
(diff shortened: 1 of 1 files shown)


strefler:
I'm not sure I'm the right person to review this, I have never really thought about this issue. Maybe rather assign Gunnar?

danielhuppmann:
@gidden @jkikstra, do you agree that fugitive vs. combustion emissions should be added in this way?

Note that there are also aggregate fugitive vs. combustion variables at the energy-supply-level, ie
> Emissions|{Level-2 Species}|Energy|Supply|Combustion
> Emissions|{Level-2 Species}|Energy|Supply|Fugitive

gidden:
I see no problem with providing a fuel-specific breakdown of combustion and fugitive emissions if modelling teams have that level of detail

jkikstra:
I also don't have an issue with it. 

Originally wasn't planning on adding this as it doesn't seem necessary, but having the freedom to play around with it later may prove to be informative.

danielhuppmann:
Rebased and bumped the tier to 3 for the combustion-vs-fugitive split as it doesn't seem to be a high priority.

PR #210, opened 2024-11-13
Add validation-criteria from AR6 WG3 Annex 3 Table 11

danielhuppmann:
This PR adds the validation-criteria from AR6 WG3 Annex 3 Table 11 for emissions and energy-production in 2020.

Table 11 also has a row on combined solar and wind generation (8.51 EJ in 2020), which I have not included because there is no standard variable for "solar plus wind" in common-definitions.

@byersiiasa, is there a reason why you combined solar and wind, and do you have the disaggregated reference values so that we could use that for more fine-grained validation criteria?

FYI @IAMconsortium/common-definitions-coordination @IAMconsortium/common-definitions-emissions @IAMconsortium/common-definitions-energy


byersiiasa:
Sorry, didnt see earlier. I think the reasoning around this was that wind and solar separately would likely cut too many scenarios and that if teams got at least the combination of the two in the right ballpark, this was enough.

PR #215, opened 2024-11-16
Implement proper tiers for final-energy variables

### LINKED ISSUE #212: Final Energy Commercial Water
The following variables are present in the ScenarioMIP template:
```
Final Energy|Commercial|Water|Desalination
Final Energy|Commercial|Water|Groundwater Extraction
Final Energy|Commercial|Water|Surface Water Extraction
Final Energy|Commercial|Water|Transfer
```
but the `Water` pattern category is missing (`Final Energy|Commercial|Water`).

Might be worth adding it.

danielhuppmann:
Per a nudge by @gunnar-pik, this PR uses a new nomenclature-feature to assign different tiers from tag-lists.

The tier is defined in the variable - if a tag-list is used for that variable, the tag-item can increment the variable-tier using `^1` or `^2`. This way, final-energy variables at a "high" aggregate, e.g *Final Energy|Agriculture|Electricity* keep the tier-1 assignment, but "higher-resolution" variables like *Final Energy|Non-Energy Use|Non-Ferrous Metals|Liquids|Electricity* have a higher tier-value.

See the attached file for a simple view of the resulting tier-assignment.
[final-energy-with-tier.xlsx](https://github.com/user-attachments/files/17784067/final-energy-with-tier.xlsx)

closes #212

### DIFF
--- a/definitions/variable/energy/final-energy.yaml
+++ b/definitions/variable/energy/final-energy.yaml
@@ -169,6 +169,10 @@
     navigate: Final Energy|Bunkers|{Transportation Sector}|{Bunker Sector}
 
 # additional variables for the water sector
+- Final Energy|Commercial|Water:
+    description: Energy consumption for water desalination, extraction, and transfer
+    unit: EJ/yr
+    tier: 2
 - Final Energy|Commercial|Water|Desalination:
     description: Energy consumption for water desalination
     unit: EJ/yr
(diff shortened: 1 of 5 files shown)


danielhuppmann:
Unless there are any objections to this PR implementing proper tiers for final-energy variables, I will merge it by end of this week.

PR #217, opened 2024-11-19
Carbon Management: Rename "Other" to "Other Source"

strefler:


danielhuppmann (commented):
Would it make sense to not replace "Other" with a more description name (e.g., "Other Method") to avoid ambiguity later? Maybe we can establish the convention that plain "Other" is only used for energy carrier?

strefler:
> Would it make sense to not replace "Other" with a more description name (e.g., "Other Method") to avoid ambiguity later? Maybe we can establish the convention that plain "Other" is only used for energy carrier?

Hm, possible. I thought it was important for carbon capture, because there are other sources and other sinks. For land and ocean I think we could also live with other to avoid inventing new terms too often

kriegler100:
As far as I remember, there is a logic behind using the generic terms "Other" and "Other Source". "Other" is for destination, e.g. energy carrier, and "Other Source" is for origin. I think there is also "Other Sector" for sectoral breakdown. Strongly suggest to stay with these generic terms in keeping with the naming convention/logic. So if it is about an other source of captured carbon, I think "Carbon Capture|Other Source" is the right choice.

kriegler100:
Sorry, need to correct myself. As far as I can see, "other source" is not used at all in the common reporting as it stands. So let us not introduce it here. I suggest to close this issue without any change.

strefler:
For Carbon Capture we need it, because we disaggregate both sources and sinks. So we need to be clear here that what we mean is "Other Source". Suggest to merge this  PR as is.

danielhuppmann (changes requested):
A few comments regarding the descriptions of the "Other" methods

strefler:
> @strefler, should I go ahead and implement the suggested updates to the descriptions as suggested and then merge this PR? Or do you want to discuss that further before I merge?

Yes, I'd say let's go ahead and merge. Thanks!

danielhuppmann (approved):

PR #219, opened 2024-11-19
Add final-energy prices by fuel (not sector)

danielhuppmann:
This PR adds final-energy prices by fuel (not including the sector). per discussion with @volker-krey and @OFR-IIASA.

### DIFF
--- a/definitions/variable/energy/energy-prices.yaml
+++ b/definitions/variable/energy/energy-prices.yaml
@@ -20,8 +20,27 @@
     weight: Secondary Energy|{Secondary Fuel Level 2}
     notes: This price should include the effect of carbon prices.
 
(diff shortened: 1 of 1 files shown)


masahirosugiyama:
@danielhuppmann Thanks for this. 

I wonder if the definition and notes are consistent or not. The definition says it should NOT include taxes (such as carbon tax?) but the note says it SHOULD include carbon tax. Can we move the notes part to the definition? 

Now, for Price|Final Energy|Electricity: 
    definition: Electricity price including transmission and distribution costs, not including taxes
    notes: This price should include the effect of carbon prices.

An alternative would be: 
    definition: Electricity price including transmission and distribution costs,
      and the effects of carbon tax and its equivalents but excluding other taxes

Issue #227, opened 2024-12-03
Emissions variable Domestic Aviation - add back in


jkikstra:
Somehow, the Domestic Aviation emissions variable did not make it in the previous pull request. 

This was previously available under `Emissions|*|Energy|Demand|Transportation|Aviation`.

For clarity, I propose to reintroduce it as `Emissions|*|Energy|Demand|Transportation|**Domestic** Aviation` for clarity.

(thx @merfort for pinging me on this; @IAMconsortium/common-definitions-emissions )

jkikstra:
OK, I see. 

But then this can only be on the global/international level, correct?

You only have global aviation? 
Does the regional 'Transportation' variable include or exclude emissions from domestic aviation?

jkikstra:
Would it be an issue for you @merfort if we don't add that aggregate variable in the reporting template? 
See https://github.com/IAMconsortium/common-definitions/pull/228#issuecomment-2522490286 .

@danielhuppmann ; thing it, we will need to create it still for the CMIP7 harmonization anyhow, as we'll need to align with the CEDS variable.

merfort:
Thanks! It is fine for now. I will come back to it later.

danielhuppmann:
About reporting for international bunkers (emissions and fuel), these values are still available at national resolution, just not consistently counted to national statistics.

Whether a model represents international bunkers at the model-native regional resolution or only at global level is a different question.

PR #228, opened 2024-12-03
Add Domestic Aviation variable

jkikstra:
Addresses #227.

@danielhuppmann we need this domestic aviation variable for CMIP7. 
But not necessarily any other differentiation. 
Should we add an "Other" category at the same level, or are there more categories that you suggest to add at this level?

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -340,6 +340,11 @@
       excluding pipeline emissions (IPCC category 1A3ei)
     unit: "{Level-2 Species}"
     tier: 2
+- Emissions|{Level-2 Species}|Energy|Demand|Transportation|Domestic Aviation:
+    description: Emissions of {Level-2 Species} from domestic aviation, being all aviation not reported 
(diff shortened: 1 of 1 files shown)


danielhuppmann:
For consistency with the energy consumption, we should also add "Domestic Shipping" - and maybe other categories including Bus, Light Duty Vehicle, Truck, Rail...

jkikstra:
> For consistency with the energy consumption, we should also add "Domestic Shipping" - and maybe other categories including Bus, Light Duty Vehicle, Truck, Rail...

Pro: consistency with energy consumption categories in the template.

Con: we don't need it for ScenarioMIP, so could create more work / some confusion on what is required until we have clear labels for ScenarioMIP (and climate emulator)?

What do you think is best @danielhuppmann

danielhuppmann:
Closing in favor of #234 to address the missing units.

Also, I'm not sure if we really need distinction of transportation-emissions between freight and passenger, and an aggregate aircraft-energy-demand variable might cause more confusion than it helps - so leaving those out for now. We can address that if we see a need for those.

PR #229, opened 2024-12-05
Fix AIM "Rest of Europe" R10 region mapping

dc-almeida:
As requested by Shinichiro Fujimori


danielhuppmann (approved):
FYI @shinichirofujimoriKU, thanks @dc-almeida!

PR #232, opened 2024-12-05
Add variables for road transportation service demand

danielhuppmann:
Per request by @gunnar-pik, this PR adds variables for energy service demand related to road transportation.

After a brief discussion with @vruijven, this PR replaces abbreviations with more descriptive terms to be more accessible to a wider audience. A mapping from NAVIGATE to the new form is included.


gunnar-pik:
I spotted a small typo: it should  be  `Fuel-Cell-Electric`, not `Fuel-Cell-ELectric`
Looks very good otherwise, thanks!

PR #235, opened 2024-12-07
Add variables for sales and stocks of vehicles

danielhuppmann:
This PR implements variables for vehicle stocks and sales based on a list sent by @gunnar-pik by email.

I did not quite understand the logic in the tier structure, and I'm skeptical if all teams can report that kind of information, hence I did not assign tier 1 (equivalent to required reporting). Instead, I assigned all variables as tier 2 except for the subtypes (car-size, two-wheeler), which are increased to 3 (using the "^1" notation).

fyi @IAMconsortium/common-definitions-transport @IAMconsortium/common-definitions-coordination


gunnar-pik:
thanks, this looks good!

PR #226, opened 2024-12-02
definition of wetland emissions

flohump:
Updated definition of wetland emissions

PR #236, opened 2024-12-09
Minor update AFOLU Emissions Land Use and Land-Use Change

flohump:
This PR adresses the issue raised by @jkikstra here:
https://github.com/IAMconsortium/common-definitions/pull/226#issuecomment-2527932422

Furthermore, this PR renames 
`Emissions|{Level-3 Species}|AFOLU|Land|Other Land Use and Land-Use Change`
to 
`Emissions|{Level-3 Species}|AFOLU|Land|Land Use and Land-Use Change`
because there is already `Emissions|{Level-3 Species}|AFOLU|Land|Other` which includes `Emissions and removals from land that cannot be accommodated in other categories `


danielhuppmann (approved):
Looks good, thanks!

Issue #224, opened 2024-12-02
Gross CO2 removals missing in emission variables


flohump:
There are variables for gross CO2 emission and carbon dioxide removal but gross CO2 removals are currently missing in emission variables. I discussed with @merfort that it would be of high value to add gross CO2 removals including sub-categories espcially for AFOLU.
This would allow to analyze how gross CO2 emissions and gross CO2 removals add-up to net CO2 emissions. With the CDR variables this is not possible for two reasons:
- not all negative CO2 emissions are considered as CDR, e.g. regrowth of vegetation on abandoned land or regrowth of timber plantations for wood production
- the CDR variables are defined as positive 

@jkikstra @IAMconsortium/common-definitions-emissions @IAMconsortium/common-definitions-land 
What do you think about this proposal?

strefler:
Agree! We prioritized net removals for the first step, but I think it would be valuable to disaggregate those into net removals and remaining emissions associated with CDR technologies. We should probably discuss the concrete structure, and also clear definitions to avoid confusion as to where emissions should be reported.

danielhuppmann:
I noticed that we now have almost-duplicate variables

```
- Carbon Removal:
    description: Gross removals of carbon dioxide (CO2) from atmospheric origin
      or biomass through deliberate human activities
    unit: Mt CO2/yr
```
introduced via #69 

```
- Gross Removals|CO2:
    description: Gross removals of carbon dioxide (CO2) excluding positive emissions,
      including removals from bioenergy with CCS (BECCS), direct air carbon capture
      and agriculture, forestry and other land use (AFOLU)
```
introduced via #238

Looking at SoCDR Edition 3 scenarios, models report both but with different values (and sometimes different sign).

Any explanation?

gidden:
As @danielhuppmann has pointed out, some teams are reporting `Gross Removals` as negative numbers (see image).

<img width="1888" height="1291" alt="Image" src="https://github.com/user-attachments/assets/eff9646b-618f-4a7f-97ab-71d12819ae91" />

[...]

Issue #222, opened 2024-11-27
Carbon Removal: Include negative emissions from regrowth of timber plantations?


flohump:
The following question came up in the ScenarioMIP land sub-group:
Should `Carbon Removal` include negative emissions from regrowth of timber plantations?
Timber plantations will be harvested after the end of the rotation period. Thus, all carbon sequestred during growth will be eventually released gain.

The question is if this is in line with the definition of `Carbon Removal` here:
https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/emissions/carbon-removal.yaml#L23
`Gross removals of carbon dioxide (CO2) from atmospheric origin or biomass through deliberate human activities`
https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/emissions/tag_land-removal-options.yaml
includes `Forest Management: description: improved forest management`

Forestry is certainly a "deliberate human activitiy". But the goal is timber production and not carbon sequestration. 
Therefore, I would argue that negative emissions from regrowth of timber plantations should not be included in `Carbon Removal`.

Does anyone agree or disagree with this @IAMconsortium/common-definitions-land @IAMconsortium/common-definitions-emissions @jkikstra @MathijsHarmsenPBL @strefler

kanishkan91:
Hello,

Thanks for the question. I'm adding my two cents here - In my opinion,
while the goal of timber plantations is indeed not carbon removal, under
reforestation scenarios (carbon price scenarios for example0 we would get
an increase in both plantations and reduction in harvests of older forests.
Thus the same plantations would be sequestering carbon by design (this time
by adjusting rotations and increasing plantations). Therefore the intent
matters less and can change based on the forestry response to a scenario?

Also, this gets a bit more complicated when thinking beyond the forestry
sector. Recently, with the increase in CCUS applications (underground
storage), there has also been an increase in enhanced oil recovery using
CCUS in oil fields. So, even though the CCS is technically being used to
recover some oil, in reality it still remains CCS. I think the same logic
can be extended to the forestry sector?

Thus,

kanishkan91:
Hello,

Thanks for the question. I'm adding my two cents here - In my opinion,
while the goal of timber plantations is indeed not carbon removal, under
reforestation scenarios (carbon price scenarios for example0 we would get
an increase in both plantations and reduction in harvests of older forests.
Thus the same plantations would be sequestering carbon by design (this time
by adjusting rotations and increasing plantations). Therefore the intent
[...]

PR #238, opened 2024-12-10
added gross removals in emissions

flohump:
This PR adds two variables 
- Gross Removals|CO2
- Gross Removals|CO2|AFOLU

Together with `Gross Emissions|CO2` and `Gross Emissions|CO2|AFOLU` this allows to understand how net emissions emerge. 

This PR is linked to issues #224 and #222 

For AFOLU, we could add more subcategories.

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -28,7 +28,14 @@
       or agriculture, forestry and other land use (AFOLU)
     unit: Mt CO2/yr
     tier: 1
-    notes: For net emissions (accounting for negative emissions), see "Emissions|CO2".
+    notes: net emissions "Emissions|CO2" = "Gross Emissions|CO2" + "Gross Removals|CO2"
+- Gross Removals|CO2:
+    description: Gross removals of carbon dioxide (CO2) excluding positive emissions. 
+      This variable includes removals from bioenergy with CCS (BECCS), direct air carbon capture
+      or agriculture, forestry and other land use (AFOLU)
+    unit: Mt CO2/yr
+    tier: 1
+    notes: net emissions "Emissions|CO2" = "Gross Emissions|CO2" + "Gross Removals|CO2"
 
 
 - Emissions|{Level-2 Species}|AFOLU:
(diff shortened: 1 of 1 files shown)


danielhuppmann (approved):
Looks good, may have some suggestions for more concise docs and validation-syntax (net emissions = gross emissions plus gross removals) in a follow-up PR.

strefler:
Ahhh, I think this was a bit fast! I would much rather have (net) CDR = gross CDR + CDR-related emissions (e.g. from calcination process for OAE). Not sure if that also works for AFOLU, but then net CO2 = gross CO2 + net CDR, not gross CDR

flohump:
> Ahhh, I think this was a bit fast! I would much rather have (net) CDR = gross CDR + CDR-related emissions (e.g. from calcination process for OAE). Not sure if that also works for AFOLU, but then net CO2 = gross CO2 + net CDR, not gross CDR

This PR is about gross removals and not CDR specifically. Of course CDR (as a negative term) is a part of gross removals. But gross removals especially in AFOLU / Land are much broader than intentional CDR. 
By adding gross removal variables, this becomes more transparent and improves clarity. 

Side note: Your sugested calcuation `(net) CDR = gross CDR + CDR-related emissions` does not work because CDR is defined as positive.

merfort:
So in order for `"Emissions|CO2" = "Gross Emissions|CO2" + "Gross Removals|CO2"` to be true, that would mean that all positive emissions (such as the emissions from calcination for OAE) would need to be part of `Gross Emissions|CO2`, right?

strefler:
> > Ahhh, I think this was a bit fast! I would much rather have (net) CDR = gross CDR + CDR-related emissions (e.g. from calcination process for OAE). Not sure if that also works for AFOLU, but then net CO2 = gross CO2 + net CDR, not gross CDR
> 
> This PR is about gross removals and not CDR specifically. Of course CDR (as a negative term) is a part of gross removals. But gross removals especially in AFOLU / Land are much broader than intentional CDR. By adding gross removal variables, this becomes more transparent and improves clarity.

I fully support having gross removals as a variable, especially for AFOLU. I'm just questioning the naming. I'm not sure a new category "Gross removals" is the right way forward, though I have to admit that I haven't thought it through yet. Maybe a disaggregation as part of Emissions|CO2|AFOLU could work, but I'm not sure. I think we would have to discuss this in the emissions subgroup.

flohump:
`Gross removals` is in line with the terminology used in GCB 2024
https://essd.copernicus.org/preprints/essd-2024-519/

Let me know if you need me in the discussions with @IAMconsortium/common-definitions-emissions

PR #248, opened 2024-12-19
Add copper tag in energy sector tag list

macflo8:
This PR adds `Final Energy|Industry|Non-Ferrous Metals|Copper*` variables.


danielhuppmann (approved):
Thanks!

PR #253, opened 2025-01-16
update definiton agricultural residues

flohump:
Clarification of definition:
Only harvested residues should be reported.


jonathandoelman (approved):
Fully agree

danielhuppmann (approved):
Thanks!

PR #259, opened 2025-01-27
Add Curacao and Sint Maarten

jkikstra:
Add two countries that in 2010, like Aruba in 1986, separated from the Netherlands to become a constituent country inside the Kingdom of the Netherlands

For more, see https://en.wikipedia.org/wiki/Kingdom_of_the_Netherlands 

This is a small PR, but points to larger questions (which could be transferred into a new Issue, i think): 
(a) when to add new countries to the mapping?
(b) is there a way to ensure that there is no overlap?
(c) is there a way to ensure completeness?
(d) should we force IAM mappings to be up-to-date with the latest common R5/R9/R10 mapping?

**How to review?**

- [ ] check if this would need to be added somewhere else
- [ ] check if this is in line with the geojson data

### DIFF
--- a/definitions/region/common.yaml
+++ b/definitions/region/common.yaml
@@ -71,7 +71,7 @@
           Ecuador, El Salvador, French Guiana, Grenada, Guadeloupe, Guatemala,
           Guyana, Haiti, Honduras, Jamaica, Martinique, Mexico, Nicaragua, Panama,
           Paraguay, Peru, Saint Kitts and Nevis, Saint Lucia,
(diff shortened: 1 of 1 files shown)


phackstock (approved):
Looks good to me.

danielhuppmann:
@lisahligono, would it be a lot of work to re-run the R5/R9/R10 geojsons for the Scenario Explorer map component with the updated country-mapping?

lisahligono:
@danielhuppmann no it's not a lot of work. I have a working script for this

PR #260, opened 2025-01-27
Update AIM_3.0.yaml

jkikstra:
Suggestion for @shinichirofujimoriKU to look at.


jkikstra:
In a bilateral call today, @shinichirofujimoriKU agreed that these suggested changes are good to go.

danielhuppmann (approved):

PR #262, opened 2025-01-27
Update MESSAGEix-GLOBIOM_2.1-R12.yaml

jkikstra:
Suggestion for @danielhuppmann to look at.


jkikstra:
Oliver agreed this is good to go ahead,

jkikstra:
I just see now that in the meantime, other versions have been added too - which are missing these updates; #283 #282.
Should probably align before merging.

PR #264, opened 2025-01-27
Update COFFEE_1.5.yaml

jkikstra:
Suggestion for @lbbaptista to look at.


jkikstra:
In a bilateral conversation, @lbbaptista and I agreed this is good to go ahead

PR #266, opened 2025-01-29
Update REMIND-MAgPIE_3.4-4.8.yml

jkikstra:
Suggestion for e.g. @merfort to look at.


merfort:
Since Kosovo is not part of the ISO country list, we don't have it in our internal region mappings, and usually map input data to Serbia, afaik. 
So I guess it's fine to add it here to Non-EU28 Europe.

jkikstra:
> usually map input data to Serbia

Yeah in general this was my assumption at first, too.
However, the data providers of GDP and Population SSP data told me that Kosovo is NOT part of Serbia. 
They're working on adding it in the next version.
So even if not an internationally recognised ISO, with data there, I think it's probably useful to have it as an additional territory with ISO in our mappings?

merfort:
Ah, interesting. Yes, then I guess it makes sense to add that to the list of countries.
But shouldn't we wait then, until the SSP update is there?
@johanneskoch94, maybe interesting for you.

jkikstra:
> But shouldn't we wait then, until the SSP update is there? 

For the aggregation of CEDS (&GFED) emissions I already need to decide where emissions from Kosovo go (i.e. to which REMIND region should I add it when making the history-vs-scenario comparison plots, and similar for harmonizing a regional pathway regions). 
Therefore I'm trying to already get them in to the region that is most applicable.

merfort:
Thanks! Makes sense. Then putting it to `Non-EU28 Europe` is definitely the right choice.

danielhuppmann (approved):
Thank you!

PR #272, opened 2025-02-07
Rename countries and units using yaml configs

lisahligono:
This pull request addresses issues discussed with @danielhuppmann  regarding the handling of special characters in region and unit names.

#### Changes

Updates to region names:
* [`country-cleanup.yaml`]: Corrects special characters in country names. For example:
  - `"C?te d'Ivoire"` is renamed to `"Côte d'Ivoire"`

Updates to unit names:
* [`unit-cleanup.yaml`]: Corrects special characters in unit names. For example:
  - `"?C"` is renamed to `"°C"`


legacy/country-cleanup.yaml
```diff
@@ -0,0 +1,5 @@
+dimension: region
+rename:
+  - "C?te d'Ivoire": "Côte d'Ivoire"
+  - "Cura?ao": "Curaçao"
+  - "R?union": "Réunion"
```
danielhuppmann, line 5: Add empty line
```suggestion
  - "R?union": "Réunion"

```

danielhuppmann (approved):
One minor comment, then good to go - thanks!

PR #273, opened 2025-02-11
subcategories for Intentional and Unintentional Gross AFOLU Removals

flohump:
This PR adds subcategories for Intentional and Unintentional Gross AFOLU Removals
- Gross Removals|CO2|AFOLU|Intentional
- Gross Removals|CO2|AFOLU|Unintentional

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -70,8 +70,27 @@
       after wood harvest or regrowth of timber plantations
     unit: Mt CO2/yr
     tier: 1
+    components:
+      - Gross Removals|CO2|AFOLU|Intentional
+      - Gross Removals|CO2|AFOLU|Unintentional
     notes: Removals reported in this category represent fluxes from the atmosphere to the land pool.
       Net emissions 'Emissions|CO2|AFOLU' = 'Gross Emissions|CO2|AFOLU' + 'Gross Removals|CO2|AFOLU'.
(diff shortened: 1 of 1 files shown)


merfort (approved):
Thanks for adding! Looks good to me.

danielhuppmann (approved):
Looks good, thanks.

merfort:
@jkikstra This is important probably for the emissions for harmonization. 
I think it would make sense to report AFOLU CO2 emissions to ESMs in these three subcategories:
`Gross Removals|CO2|AFOLU|Intended`  (which should basically be `Carbon Removal|Land Use|Re/Afforestation`)
`Gross Removals|CO2|AFOLU|Unintended`
`Gross Emissions|CO2|AFOLU`

such that they can distinguish between (intended) CDR from afforestation and between additional gross negative and gross positive emissions. Given that ESMs will model the land carbon fluxes endogenously, they will probably not use that information directly, but I think it is helpful for comparison.

flohump:
`Gross Removals|CO2|AFOLU|Intended` should be the same as `Carbon Removal|Land Use` but with different sign. `Carbon Removal|Land Use|Re/Afforestation` is likely the largest component but also other land-based CDR options are included.

merfort:
To avoid confusion: it is `Gross Removals|CO2|AFOLU|Intentional` / `Gross Removals|CO2|AFOLU|Unintentional` (and not `...|Intended` / `...|Unintended`)

Issue #270, opened 2025-02-05
Add Useful Energy variables to the template


jkikstra:
As discussed in energy breakout of ScenarioMIP today. 
Ping amongst others @gunnar-pik @vruijven 

Requirements:
* Complete set 
* Key sectors for assessing efficiency
* Clear guidance on _how to report_ / what unit.

Should include at least: 
* Useful Energy|Transportation
* Useful Energy|Residential and Commercial
* Useful Energy|Industry

Probably better full list: directly the same as the Final Energy tree (which is not having the full 'components' list currently): https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/energy/final-energy.yaml

jkikstra:
So OK, my proposal for now is this.

Maybe @OFR-IIASA and/or @khaeru weigh in if you would think that there is a better way?

We follow:
(a) the sectoral structure of Final Energy (only top level), and adopt its definitions 
(b) report in fuel-equivalent (which we add to the variable description, or the `note`)


variable | description | unit | tier
-- | -- | -- | --
Useful Energy | Useful energy consumption by all end-use sectors and all fuels, including   non-energy use, excluding transmission/distribution losses | EJ/yr | 1
Useful Energy (w/o bunkers) | Useful energy consumption by all end-use sectors and all fuels including   non-energy use, excluding international aviation and shipping (see 'Useful Energy\|Bunkers\|*') and excluding transmission/distribution losses | EJ/yr | 2
Useful Energy\|Agriculture | Useful energy consumption by the agriculture sector including fishing | EJ/yr | 1
Useful Energy\|Bunkers | Useful energy consumption by the international aviation and shipping | EJ/yr | 1
Useful Energy\|Carbon Management | Total energy use for carbon management, i.e., capture and/or removal of   CO2 | EJ/yr | 2
Useful Energy\|Commercial | Useful energy consumption by the commercial sector | EJ/yr | 1
Useful Energy\|Industry | Useful energy consumption by the industrial sector excluding non-energy   use (e.g.feedstocks) | EJ/yr | 1
Useful Energy\|Non-Energy Use | Useful energy consumption in non-combustion processes | EJ/yr | 2
Useful Energy\|Other Sector | Useful energy consumption by the other sectors | EJ/yr | 2
Useful Energy\|Residential | Useful energy consumption by the residential sector | EJ/yr | 1
[...]

PR #275, opened 2025-02-13
Update energy service variables & Add Useful Energy

jkikstra:
For ScenarioMIP analysis, per discussion in #270 and also discussion with @vruijven @christophbertram @gunnar-pik @volker-krey

### DIFF
--- a/definitions/variable/sdg-indicators/sdg.yaml
+++ b/definitions/variable/sdg-indicators/sdg.yaml
@@ -170,31 +170,6 @@
     unit: km3/yr
     sdg: 6
     tier: 2
-- Useful Energy|Transportation|Passenger [per capita]:
-    description: Useful energy per capita for passenger transport
-    unit: GJ/cap/yr
-    sdg: 7
-    weight: Population
-- Useful Energy|Industry [per capita]:
-    description: Useful energy per capita for industry
-    unit: GJ/cap/yr
-    sdg: 7
-    weight: Population
-- Useful Energy|Residential and Commercial [per capita]:
-    description: Useful energy per capita for buildings
-    unit: GJ/cap/yr
-    sdg: 7
-    weight: Population
-- Useful Energy|Residential [per capita]:
-    description: Useful energy per capita for buildings
-    unit: GJ/cap/yr
-    sdg: 7
-    weight: Population
-- Useful Energy|Commercial [per capita]:
-    description: Useful energy per capita for buildings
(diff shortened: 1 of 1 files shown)


jkikstra:
Not sure what exactly is causing the 'illegal character' error:

```
=========================== short test summary info ============================
FAILED tests/test_legacy.py::test_legacy_variables - ValueError: Collected 2 errors:
  1. Unexpected character in variable: 'Production|Iron and Steel'. Check for illegal characters and/or if tags were spelled correctly.
  2. Unexpected character in variable: 'Production|Non-Metallic Minerals'. Check for illegal characters and/or if tags were spelled correctly.
FAILED tests/test_reserved_terms.py::test_variable_ops_as_square_brackets - ValueError: Collected 2 errors:
  1. Unexpected character in variable: 'Production|Iron and Steel'. Check for illegal characters and/or if tags were spelled correctly.
  2. Unexpected character in variable: 'Production|Non-Metallic Minerals'. Check for illegal characters and/or if tags were spelled correctly.
========================= 2 failed, 1 warning in 6.05s =========================
```

definitions/variable/industry/production.yaml
```diff
@@ -1,14 +1,28 @@
+- Production|Iron and Steel:
+    description: Production of Iron and Steel
+    unit: Mt/yr
+    tier: 1
+    components:
+        - Production|Iron and Steel|{Iron Commodity}
```
danielhuppmann, line 6: This replacement doesn't work (if the tag isn't used in the variable name), hence the illegal-character error.

jkikstra: OK, so if I want to add a top-level variable like this & add the list of components, would there be another way, or would I always have to list them all, or just not list the components?

danielhuppmann: I would advise against listing components if these are all sub-categories. This is more confusing than helpful.

jkikstra: OK, removed.

jkikstra:
OK thanks @vruijven for your thoughts. Now updated. Would be great if we can move this in such that we can use it for JUSTMIP submissions.

jkikstra:
@phackstock @danielhuppmann can we go ahead here, or is there any additional changes needed?

Note: we could also delete Useful Energy [per capita] variables from the sdg YAML

danielhuppmann (approved):
Looks good to me, thanks!

PR #276, opened 2025-02-14
Add note on bunker emissions.

jkikstra:
In line with [IPCC 2006 reporting guidance](https://www.ipcc-nggip.iges.or.jp/public/2019rf/pdf/1_Volume1/19R_V1_Ch08_Reporting_Guidance.pdf): 

"The summary tables also allow reporting of memo items including international bunkers and multilateral
operations. These emissions are not included in national total emissions of greenhouse gases."


strefler:
We've had quite a bit of discussion on that in the NAVIGATE project, and decided that we would like region emissions to always add up to global. Therefore also international bunkers should be added to regions. To account for any potential misunderstandings, we added total emissions without bunkers as a separate category.

jkikstra:
That's another way to solve things; like in the Final Energy tree.
But currently we don't have that in https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/emissions/emissions.yaml

PR #278, opened 2025-02-17
Add investment variables for power generation, storage and grid

danielhuppmann:
While working on scenario-compass migration, I noticed that the electricity-investment variables were still missing.


danielhuppmann:
Merging ahead of the SWG meeting

PR #280, opened 2025-02-17
Varaible From National scenario template

FlorianLeblancDr:
I went through the national scenario template to see what needed to be imported (double check welcomed).
 [https://data.ene.iiasa.ac.at/ar6-scenario-submission/static/files/national.zip](https://data.ene.iiasa.ac.at/ar6-scenario-submission/static/files/national.zip)

- Mostly Debt variables
- I found Investment|Clean Energy not well defined, which seems to be a sub-category of Investment|Energy Supply so I leave it behind
- Public Consumption is redundant with Expenditure|Government, which is a better name, so I leave it behind
- Labour Supply is defined under "Labor Force" now
- remaining: Subsidies, Taxes, Trade variables which should come in a dedicated PR

### DIFF
--- a/definitions/variable/macro-economy/gdp.yaml
+++ b/definitions/variable/macro-economy/gdp.yaml
@@ -6,6 +6,10 @@
     description: GDP converted to USD using purchasing power parity (PPP)
     unit: billion USD_2010/yr
     tier: 1
+- Value Added:
+    description: Total value added
+    unit: billion USD_2010/yr
+    tier: 1
 - Value Added|{Economic Sectors}:
     description: Value added by the {Economic Sectors}
     unit: billion USD_2010/yr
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Looks good to me, any comments from the other @IAMconsortium/common-definitions-macro-economy team colleagues?

lolow:
That's good for me as well

danielhuppmann:
While preparing for the IAMC-SWG meeting on Friday, I remembered the comment about too many tier-1 variables, so I'm wondering if these variables really have to be categorized as tier 1 (in the sense of "a scenario cannot be analyzed without it, strictly required info"). Please consider downgrading to tier 2.

FlorianLeblancDr:
You right, done.
I also moved Capital Stock to tier 2: to me this is something every model should have but I don't want to be too restrictive. @lolow any idea on this?

About friday, Daniel: sorry for being late, this act as a deadline for us to have at least a bunch of PR ready to reframe old templates, but we might need a few days more to get it done.

PR #283, opened 2025-02-18
Add region definition and mapping for "MESSAGEix-GLOBIOM-GAINS 2.1-R12"

danielhuppmann:
This PR adds a new model-region-group "MESSAGEix-GLOBIOM-GAINS 2.1-R12" and the mapping for the "M"-variant of that model.


OFR-IIASA (approved):
lgtm

PR #284, opened 2025-02-19
Units: add consistency in t DM units

klau506:
Standardize the notation for `t DM` units to ensure consistency, as both `tDM` and `t DM` were previously used.


danielhuppmann (approved):
Nice, thank you!

PR #289, opened 2025-02-19
Price carbon

FlorianLeblancDr:
Same as defined and discussed in #38

### DIFF
--- a/definitions/variable/macro-economy/price.yaml
+++ b/definitions/variable/macro-economy/price.yaml
@@ -22,7 +22,7 @@
        can result in counter-intuitive aggregation results when used as weights for
        the region-aggregation. Also the [weighted by Final Energy] methods is
        complementary and overcomes other aggregation issues - regions with high carbon
-       prices can be associated with low emission levels, meaning the its contribution
+       prices can be associated with low emission levels, meaning its contribution
        to the emission-weighted average of carbon prices will be low, regardless
        of the share of the global energy system the carbon price is applied to.
 - Price|Carbon [weighted by Final Energy]:
--- a/definitions/variable/macro-economy/price.yaml
+++ b/definitions/variable/macro-economy/price.yaml
@@ -19,7 +19,12 @@
     unit: USD_2010/t CO2
     skip-region-aggregation: true
     notes: Regional CO2 emissions can turn negative at different points in time, which
-       can result in counter-intuitive timeseries when used as weights for averages.
+       can result in counter-intuitive aggregation results when used as weights for
+       the region-aggregation. Also the [weighted by Final Energy] methods is
+       complementary and overcomes other aggregation issues - regions with high carbon
+       prices can be associated with low emission levels, meaning the its contribution
+       to the emission-weighted average of carbon prices will be low, regardless
+       of the share of the global energy system the carbon price is applied to.
 - Price|Carbon [weighted by Final Energy]:
     description: Price of carbon (weighted by final energy consumption across regions)
     unit: USD_2010/t CO2
--- a/definitions/variable/macro-economy/price.yaml
+++ b/definitions/variable/macro-economy/price.yaml
@@ -2,25 +2,25 @@
     description: Price of carbon (as reported by the model)
     unit: USD_2010/t CO2
     region-aggregation:
-      - Price|Carbon (mean):
+      - Price|Carbon [Mean]:
           method: mean
-      - Price|Carbon (weighted by Emissions|CO2):
+      - Price|Carbon [weighted by Emissions|CO2]:
           weight: Emissions|CO2
           drop_negative_weights: false
-      - Price|Carbon (weighted by Final Energy):
+      - Price|Carbon [weighted by Final Energy]:
           weight: Final Energy
     check-aggregate: false
-- Price|Carbon (mean):
+- Price|Carbon [Mean]:
     description: Price of carbon (computed as unweighted mean across regions)
     unit: USD_2010/t CO2
     skip-region-aggregation: true
-- Price|Carbon (weighted by Emissions|CO2):
+- Price|Carbon [weighted by Emissions|CO2]:
     description: Price of carbon (weighted by regional CO2 emissions)
     unit: USD_2010/t CO2
     skip-region-aggregation: true
-    note: Regional CO2 emissions can turn negative at different points in time, which
-          can result in counter-intuitive timeseries when used as weights for averages
-- Price|Carbon (weighted by Final Energy):
+    notes: Regional CO2 emissions can turn negative at different points in time, which
+       can result in counter-intuitive timeseries when used as weights for averages.
+- Price|Carbon [weighted by Final Energy]:
     description: Price of carbon (weighted by final energy consumption across regions)
     unit: USD_2010/t CO2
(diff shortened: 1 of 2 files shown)


danielhuppmann:
This is basically a copy of #38 (which I started 18 months ago) so pinging @orichters & @robertpietzcker again.

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
+- Price|Carbon:
+    description: Price of carbon (as reported by the model)
+    unit: USD_2010/t CO2
+    region-aggregation:
+      - Price|Carbon (mean):
```
danielhuppmann, line 5:
```suggestion
      - Price|Carbon [Mean]:
```

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
...
+          method: mean
+      - Price|Carbon (weighted by Emissions|CO2):
```
danielhuppmann, line 7:
```suggestion
      - Price|Carbon [weighted by Emissions|CO2]:
```

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
...
+          weight: Emissions|CO2
+          drop_negative_weights: false
+      - Price|Carbon (weighted by Final Energy):
```
danielhuppmann, line 10:
```suggestion
      - Price|Carbon [weighted by Final Energy]:
```

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
...
+          weight: Final Energy
+    check-aggregate: false
+- Price|Carbon (mean):
```
danielhuppmann, line 13:
```suggestion
- Price|Carbon [Mean]:
```

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
...
+    description: Price of carbon (computed as unweighted mean across regions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+- Price|Carbon (weighted by Emissions|CO2):
```
danielhuppmann, line 17:
```suggestion
- Price|Carbon [weighted by Emissions|CO2]:
```

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
...
+    description: Price of carbon (weighted by regional CO2 emissions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+    note: Regional CO2 emissions can turn negative at different points in time, which
+          can result in counter-intuitive timeseries when used as weights for averages
+- Price|Carbon (weighted by Final Energy):
```
danielhuppmann, line 23:
```suggestion
- Price|Carbon [weighted by Final Energy]:
```

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
...
+- Price|Carbon (mean):
+    description: Price of carbon (computed as unweighted mean across regions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+- Price|Carbon (weighted by Emissions|CO2):
+    description: Price of carbon (weighted by regional CO2 emissions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+    note: Regional CO2 emissions can turn negative at different points in time, which
+          can result in counter-intuitive timeseries when used as weights for averages
```
danielhuppmann, line 22: Use `notes`, harmonize indent and full stop at end of sentence for consistency.
```suggestion
    notes: Regional CO2 emissions can turn negative at different points in time, which
       can result in counter-intuitive timeseries when used as weights for averages.
```

danielhuppmann (commented):
A few suggestions for following a consistent style.

definitions/variable/macro-economy/price.yaml
robertpietzcker: sounds good!
I would propose to extend the notes: 
"Regional CO2 emissions can turn negative at different points in time, which can result in counter-intuitive aggregation results when used as weights for the region-aggregation. Also, regions will high carbon prices will have lower emissions, meaning the emission-weighted average of carbon prices will be low, although the high carbon price may be applied to a large share of the global energy system. Using final energy as weight overcomes these aggregation issues."

which points to the alternative without these problems, but still doesn't say "take this as default".

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,26 @@
+- Price|Carbon:
```
robertpietzcker, line 1: I am wondering - is it useful to have a "pure" variable without explicit region aggregation? 
I fear this could lead to constant need for questions back to the modelers "how did you aggregate this", and I am not sure we gain anything? 

also, what would the scenario explorer do for the automatic aggregation - doesn't it need a specification which of the following variants is applied when reading in "Price|Carbon"? 

Maybe I overlook something, but otherwise I would propose to drop this general term and only have the three explicit versions below.

robertpietzcker: Ah, so teams upload with "Price|Carbon", and then in the explorer one will see "Price|Carbon [Mean]" (or whatever was decided as aggregation mechanism by the project lead?

FlorianLeblancDr:
@robertpietzcker I just added your suggested notes with some rephrasing. 
Is it ok like this?
Should we split the note between the two variable (weighted by emissions and by final energy?). This would avoid long notes.

definitions/variable/macro-economy/price.yaml
```diff
@@ -0,0 +1,31 @@
...
+    skip-region-aggregation: true
+- Price|Carbon [weighted by Emissions|CO2]:
+    description: Price of carbon (weighted by regional CO2 emissions)
+    unit: USD_2010/t CO2
+    skip-region-aggregation: true
+    notes: Regional CO2 emissions can turn negative at different points in time, which
+       can result in counter-intuitive aggregation results when used as weights for
+       the region-aggregation. Also the [weighted by Final Energy] methods is
+       complementary and overcomes other aggregation issues - regions with high carbon
+       prices can be associated with low emission levels, meaning the its contribution
```
robertpietzcker, line 25: meaning its contribution (delete the "the")

robertpietzcker (approved):
thanks - looks good to me!

robertpietzcker:
> Should we split the note between the two variable (weighted by emissions and by final energy?). This would avoid long notes.

I have to admit I don't know if there is a preferred approach - both would be fine for me.

FlorianLeblancDr:
> > Should we split the note between the two variable (weighted by emissions and by final energy?). This would avoid long notes.
> 
> I have to admit I don't know if there is a preferred approach - both would be fine for me.

Let's keep it like this then.
@danielhuppmann This seems ready to be merged.

FlorianLeblancDr:
ping @danielhuppmann  Should we merge this?

danielhuppmann (approved):
Thank you, looks good

PR #291, opened 2025-02-24
fix species not flexible in energy emission descriptions

pweigmann:
Small fix of `Emissions|<gas>|Energy` incorrectly being described as CO2 emissions for all species.


danielhuppmann (approved):
Thank you!

PR #268, opened 2025-01-29
Update GCAM_V7.1.yaml

jkikstra:
Suggestion for e.g. @d3y419 to look at.

Not the most urgent, but good to align.

PR #257, opened 2025-01-23
Fix korea region

d3y419:
South Korea was incorrectly categorized in a few models. I went through the list and fixed them.

PR #296, opened 2025-02-26
Region definition and mappings for GCAM 6.0

danielhuppmann:
This PR copies over the addition of missing countries for GCAM 7.1 as implemented by @jkikstra in #268 (but with cleaned-up formatting) and adds the same regions and mappings for GCAM 6.0.

Pinging @ @d3y419


d3y419:
Looks good to me, except Korea still stuck in the wrong R10 per https://github.com/IAMconsortium/common-definitions/pull/257
@christophbertram can you also review that?

danielhuppmann:
Thanks for catching that @d3y419!

Issue #265, opened 2025-01-28
Treatment of embedded carbon in feedstocks


jayfuhrman:
In looking at the current carbon-management.yaml file for reporting GCAM results for the scenariomip project, there doesn't seem to be a clear way of classifying carbon embedded in feedstocks.  

Not sure how other models handle this but GCAM reports non-combustion use of refined liquids products to produce e.g., plastics + asphalt as "carbon sequestration".  However the carbon dioxide isn't actually captured for this purpose, it's just already embedded in the petroleum or gas used to produce feedstocks, and then not emitted at the manufacturing facility.  

The closest variable definition for this appears to be Carbon Capture|Utilization|Materials|Short-Lived or Carbon Capture|Utilization|Materials|Long-Lived, depending on whether the feedstock in question is used for plastics (short-lived), or asphalt (long-lived)

Alternatively, do we want to create a new variable definition (e.g., something like Carbon Capture|Feedstock Embedded Carbon|Materials|Short-Lived), or simply omit non-biomass feedstocks where there is no dedicated effort to capture a CO2 stream and incorporate it into the product, from the reporting templates?

@christophbertram @jonsampedro

fschreyer:
Thanks for bringing this up. I think, there are generally a couple of open questions on the treatment of feedstocks and their carbon flows and emissions. 

In REMIND, we report non-fossil feedstocks in long-lived products / landfills under ``Carbon Removal|Long-Lived Materials``. One purpose of introducing variables for "Embedded Carbon" could be that one could track how much carbon (of which origin) goes where in the end. Also, it could be used for consistency checks as I would suppose that emitted fossil / stored non-fossil carbon should show up as positive / negative waste emissions in the model somewhere. However, I am not sure about the priority of this level of detail right now. Maybe people with dedicated projects on chemicals and circularity can say more.

I do not entirely understand the difference between Carbon Capture|Utilization|Materials|Short-Lived and your proposal Carbon Capture|Feedstock Embedded Carbon|Materials|Short-Lived? Is it about whether or not the material was produced via CCU? In my understanding, e.g. fossil carbon in plastics made from oil should not be treated under a "carbon capture" variable because I associate carbon capture with capturing CO2 from a flue gas.

jayfuhrman:
you're right, I think it makes more sense to just have a third tier1 variable called Feedstock Embedded Carbon|Materials|Short-Lived and omit the Carbon Capture

strefler:
[...]

PR #299, opened 2025-03-03
Update carbon-management.yaml

jayfuhrman:
re: issue #265 I suggest adding a new tier1 variable to reflect fossil or biomass carbon e.g., in plastics or asphalt manufacture that is embedded in the product, rather than being emitted.  In my view only biomass-derived carbon should be reported as carbon removal.  @fschreyer

### DIFF
--- a/definitions/variable/emissions/carbon-management.yaml
+++ b/definitions/variable/emissions/carbon-management.yaml
@@ -231,3 +231,12 @@
     tier: 2
     lower_bound: 0
     navigate: Carbon Capture|Usage|{Carbon Utilization Type}
+
+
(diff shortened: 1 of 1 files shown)


danielhuppmann:
This proposed variable seems to be a duplicate of "Carbon Capture|Utilization|Materials|*".

And "Carbon Removal|*" describes "Gross removals of carbon dioxide (CO2) from atmospheric origin or biomass through deliberate human activities" (see the description).

strefler:
Agree with Daniel. Why can it not be reported in either Carbon Capture|Utilization|Materials or Carbon Removal|Materials?

PR #303, opened 2025-03-24
improved definition for Agricultural Demand|Crops|Food

flohump:
`Agricultural Demand|Crops|Food` should include demand for food crops from primary products (e.g. rice) and processed products (e.g. oil -> soybean)


jonathandoelman (approved):
I approve the new definition, however I want to note here that it does it make it much harder to align between models. Processing the FAO food demand is non-trivial and it is very unlikely that models will be able to align unless we set up a process of collectively defining the FAO data processing.

danielhuppmann (approved):

PR #304, opened 2025-03-27
Propose tier change for Ocean CDR

tmkhsgw:
Just as a minor point, I would propose change tiers from 1 to 3 for ocean CDRs that still many models could not submit.


danielhuppmann:
Makes sense to me, any objections @IAMconsortium/common-definitions-emissions or @IAMconsortium/common-definitions-coordination?

jkikstra:
Sure, Tier 2 or Tier 3 here seems to make sense, indeed not Tier 1 (given it is in its infancy, I would this does not seem essential).

PR #305, opened 2025-04-03
Update IMAGE_v3.4.yaml

VassilisDaioglou:
Corrected R10 definition of "China+" to include KOR IMAGE region.


danielhuppmann (approved):
Thanks for the fix

Issue #310, opened 2025-04-16
Countries belong to multiple regions


fabiosferra:
Dear all,

I am using the regional mapping files from `definitions/region/native_regions` and I noticed that some countries belong to different native regions. 

I made a list below:
```
Model: IMACLIM 2.0, Country: Kiribati, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Marshall Islands, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Micronesia, Federated States of, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Nauru, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Niue, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Northern Mariana Islands, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Palau, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Papua New Guinea, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Samoa, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
Model: IMACLIM 2.0, Country: Solomon Islands, Regions: ['IMACLIM 2.0|Japan, Australia, New Zealand and Korea', 'IMACLIM 2.0|Rest of Asia']
[...]

PR #311, opened 2025-04-17
fix JAN native region

FlorianLeblancDr:


danielhuppmann (approved):
Thank you for the quick fix!

closes #310

PR #314, opened 2025-04-25
ISIC sector definition for industrial production

FlorianLeblancDr:
@IAMconsortium/common-definitions-industry 
Please check if this is correct
help: [ISIC divisions](https://unstats.un.org/unsd/classifications/isic/rev5consultation)

### DIFF
--- a/definitions/variable/industry/tag_industry_sectors.yaml
+++ b/definitions/variable/industry/tag_industry_sectors.yaml
@@ -1,13 +1,21 @@
 - Industry Sector:
     - Chemicals:
-        description: production of chemicals
+        description: production of chemicals, chemical products,
+          basic pharmaceutical products and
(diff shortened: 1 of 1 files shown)


macflo8:
Hi @FlorianLeblancDr,
Thanks for adding some descriptions. This looks mostly good to me.

Just as a side note:
I came up with longer version a while ago (https://github.com/macflo8/common-definitions/blob/industry_old/definitions/variable/industry/tag_industry_sectors.yaml) that matches the IEA Energy Balances flows. But we ended up starting with a simpler version for the first common-definitions implementation. In case there is a need to match all IEA energy flows in the future, we can use it to extend the tag list.

FlorianLeblancDr:
Thanks @macflo8 
The detailed nomenclature match could be helpful indeed

@danielhuppmann this seems ready to be merged

PR #315, opened 2025-04-28
Add region-definitions for legacy REMIND[-MAgPIE] versions

danielhuppmann:
For completeness and automated downscaling for the Scenario Compass project (@fabiosferra), this PR adds region-definitions and legacy mappings of the 12-region version of previous REMIND versions used in earlier projects.


laurinks (commented):
Thank you, @danielhuppmann! Looks generally fine to me. 

However, I would suggest to add REMIND 3.0 analogously to [REMIND 3.1](https://github.com/IAMconsortium/common-definitions/blob/main/mappings/REMIND-MAgPIE/REMIND_3.1.yaml), i.e. enable submission based on 12 region version or 21 region version.

danielhuppmann:
Thanks @laurinks, good call...

PR #316, opened 2025-05-09
More specific definition of crop yields

flohump:
Crop yields can be based on physical or harvested area.
This PR refines the definition of the existing `Yield|Cropland|{Crop Types}` variables (physical) and adds new variables for yield based on harvested area `Yield|Cropland Harvested|{Crop Types}`.

### DIFF
--- a/definitions/variable/afolu/agriculture.yaml
+++ b/definitions/variable/afolu/agriculture.yaml
@@ -172,7 +172,7 @@
     weight: Land Cover|Cropland|{Crop Types}
     tier: 2
     notes: Yields in FAOSTAT are reported on wet matter basis for harvested cropland area
-- Yield|Harvested Cropland|{Crop Types}:
+- Yield|Cropland|{Crop Types} [Harvested Area]:
     description: Yield of {Crop Types}, calculated as production (dry matter) divided by harvested cropland area
     unit: t DM/ha/yr
     weight: Land Cover|Cropland|{Crop Types}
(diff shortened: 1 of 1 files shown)


jonathandoelman:
Hi Florian, all,

Thanks for creating the issue. I am not sure if it is useful to have both variants of yield represented in the common definitions template, it may actually be quite confusing. I would argue that yield on the basis of harvested area is the most common and sensible indicator. On the other hand, we also don't report harvested area per crop type, but physical area as part of the Land Cover category, even though this is a bit of a constructed variable as well. In AGMIP the focus is also for area information per crop type on harvested area.

I dont want to create a lot of changes at this late stage, but it would be good to hear what others think. For sure we need to be clear about the definition. @IAMconsortium/common-definitions-land please let us know what you think.

Jonatahn

flohump:
I think clarity is essential here. I suggested including both definitions because it helps make transparent that different interpretations of "yield" exist and are in use across modeling teams. Including both doesn't necessarily mean that all teams need to report both. Rather it can serve to highlight the distinction and encourage more precise usage. Re-defining `Yield|Cropland|{Crop Types}` as yield based on physical area doen't necessarily mean making it a default. But if most teams would indeed report yields based on harvested as default, we could of course do it the other way around, i.e. re-defining `Yield|Cropland|{Crop Types}` as yield based on harvested area, and adding a new variable for physical area. @IAMconsortium/common-definitions-land

jonathandoelman (approved):
Ok! Sounds like a decent approach for now. Also good I think not to change the original variable. Let's keep in mind that we may want to change the definition again depending on what the majority of models report.

jonathandoelman (approved):

flohump:
Thanks for the suggestion. Indeed that seems the better solution. I will adjust the PR accordingly.

danielhuppmann (approved):
Thanks, this is much less likely to lead to confusion...

PR #317, opened 2025-05-09
Cropland: add fallow land and agroforestry trees

flohump:
This PR adds sub-categories for fallow land and agroforestry trees to `Land Cover|Cropland|`

### DIFF
--- a/definitions/variable/afolu/land-cover.yaml
+++ b/definitions/variable/afolu/land-cover.yaml
@@ -22,8 +22,20 @@
     unit: million ha
     agmip: Land Cover (LAND) - All crops (CRP)
     tier: 1
+    components:
+      - Land Cover|Cropland|{Crop Types}
+      - Land Cover|Cropland|Fallow
+      - Land Cover|Cropland|Trees
 - Land Cover|Cropland|{Crop Types}:
     description: Cropland for {Crop Types}
+    unit: million ha
+    tier: 2
+- Land Cover|Cropland|Fallow:
+    description: Fallow cropland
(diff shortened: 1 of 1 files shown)


flohump:
@danielhuppmann The test stops with `Illegal character(s) '{', '}' in components of variable 'Land Cover|Cropland'.`
It seems that {} don't work in components if the brackets are at the end.
Is there a workaround?

jonathandoelman:
I think we need a bit more clarity here on the definitions.

Regarding fallow land: do you know if models represent this explicitly? For IMAGE we can say something on the difference between arable land and harvested area, but thats actually a mix between fallow and double cropping. I think we could  come up with a bit smarter analysis, but that not something for the next few weeks.

Regarding agroforestry trees, do you have in mind the actual tree/canopy cover of agroforestry trees within agricultural land? Or the agricultural land area that is defined as agroforestry, i.e. with a minimum number of trees? How do you approach this in MAgPIE? In any case we dont have this included in IMAGE at the moment.

definitions/variable/afolu/land-cover.yaml
```diff
@@ -22,10 +22,22 @@
     unit: million ha
     agmip: Land Cover (LAND) - All crops (CRP)
     tier: 1
+    components:
+      - Land Cover|Cropland|{Crop Types}
+      - Land Cover|Cropland|Fallow
+      - Land Cover|Cropland|Trees
```
danielhuppmann, line 28: The `{}` tag-lists cannot be used in lists of components (yet).

PR #320, opened 2025-06-10
secondary products - per-capita availability and intake

flohump:
Secondary products such as oils and sugar are currently missing in the per-capita food availability and intake variables. 
They are needed to reconcile differences in the top level variables reported by the differend modelling teams. 
This PR adss secondary products for per-capita availability and intake.

### DIFF
--- a/definitions/variable/afolu/food-system.yaml
+++ b/definitions/variable/afolu/food-system.yaml
@@ -28,6 +28,12 @@
     sdg: 2
     weight: Population
     tier: 3
+- Food Availability|Secondary Products [per capita]:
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Looks ok to me, waiting for @IAMconsortium/common-definitions-land to comment.

flohump:
I think we agreed to include secondary products in `Food Availability|Crops [per capita]`.
Therefore, the revised definition "Amount of food available from crops and secondary products"
In addition, I added `Food Availability|Crops|Secondary Products [per capita]` as optional variable for those teams who can report on this.
Does that make sense?

danielhuppmann (approved):

PR #322, opened 2025-06-16
Add mapping from model-native regions to R9 for REMIND-MAgPIE and REMIND

laurinks:
This PR adds mappings from model-native regions of REMIND and REMIND-MAgPIE to common regions R9.

Note that `European Union (R9)` reflects membership status as of 2023 (EU 27), but `EUR` in REMIND (12 regions) and REMIND-MAgPIE includes the United Kingdom (EU 28). In REMIND (21 regions), this can be partly improved by mapping UKI (United Kingdom and Ireland) to `Other OECD (R9)` (even though Ireland should be part of `European Union (R9)`).

As the description in `definitions/region/common.yaml` contains the note _"Depending on the spatial resolution, reporting by some models may differ from the current list of EU27 members (2023)."_, this nonetheless seems to be the best possible mapping to common regions R9.

@flohump, @strefler, @Renato-Rodrigues: Thanks for double-checking.

### DIFF
--- a/mappings/REMIND-MAgPIE/REMIND-MAgPIE_3.2-4.6.yaml
+++ b/mappings/REMIND-MAgPIE/REMIND-MAgPIE_3.2-4.6.yaml
@@ -47,6 +47,28 @@ common_regions:
   - Reforming Economies (R5):
       - REF
 
+  - European Union (R9):
+      - EUR
+  - USA (R9):
+      - USA
+  - Other OECD (R9):
+      - JPN
+      - CAZ
+      - NEU
+  - China (R9):
+      - CHA
+  - India (R9):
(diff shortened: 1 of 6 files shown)


flohump (approved):
Looks good. Thanks

laurinks:
Thanks for checking! I also got direct feedback from Robert on H21, so this should also be fine. 

@phackstock, @danielhuppmann: Is there a chance to merge this within the next few hours so that it is already applied to the resubmission later this afternoon? Thanks a lot!

laurinks:
Great, thanks! We need this for ScenarioMIP at this point, i.e. https://github.com/iiasa/ssp-submission-workflow

PR #323, opened 2025-06-26
Update GCAM_7.1.yaml to fix R5 mappings

realxinzhao:
Fix ASIA, REF, and OECD & EU

### DIFF
--- a/mappings/GCAM_7.1.yaml
+++ b/mappings/GCAM_7.1.yaml
@@ -73,7 +73,6 @@ common_regions:
             
     # R5 regions
   - Asia (R5):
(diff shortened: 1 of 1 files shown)


realxinzhao:
@christophbertram @flohump 

This commit updated the GCAM R5 mapping. 
@christophbertram  can you review this? 

I think the GCAM R5 population should be fixed later.

danielhuppmann:
PR looks ok to me but keep in mind that this will only affect new submissions to any IIASA database after this PR is merged (and only those projects that use the latest of **common-definitions** - it will not update already existing submitted scenario data.

PR #324, opened 2025-06-30
added Expenditure|Households|Agriculture [Share]

flohump:
This PR adds the following new reporting variable:
`Expenditure|Households|Agriculture [Share]`
description: Expenditure share for agricultural products in total income based on farm-gate prices (without value added from food processing chain)

This is need because most IAM team can't report `Expenditure|Households|Food [Share]`, which includes value added from the food processing chain in addition to the expenditures for agricultural products.


flohump:
I see. That makes sense. I changed the name of the variable according.

danielhuppmann (approved):

PR #332, opened 2025-07-08
Add Kosovo to REMIND native-region definitions

danielhuppmann:
This PR applies the fix by @jkikstra in #266 - adding Kosovo to the native-region definitions - to the other native-region definitions of the REMIND model.


danielhuppmann:
See this related discussion at https://github.com/IAMconsortium/common-definitions/pull/266#issuecomment-2624678608.

laurinks:
Looks good to me. However, I noted that there were some more clean-ups in https://github.com/IAMconsortium/common-definitions/pull/266 regarding duplications in EU28 (e.g. Gibraltar). I would suggest to apply these here as well.

laurinks (approved):
Thanks!

PR #335, opened 2025-07-30
Change G20 members for GCAM to native regions to comply with new formatting

dc-almeida:
For nomenclature v0.26+


dc-almeida:
Replaced by #336

PR #337, opened 2025-08-08
Add Saint Helena to ScenarioMIP models (+ POLES)

jkikstra:
Is used in the country-mapping in ScenarioMIP, so adding where it was missing.


danielhuppmann:
I guess this doesn't make much difference either way, but is this change reflecting how modelling teams actually define their native regions, or is this how you think they *should* define their regions?

jkikstra:
I don't know _what teams are precisely doing_, so it's my guess for _what I think they are doing_.

From my side, there's no right or wrong way of under which region it would fall. (so there's no _should_)

But it does need to be defined somewhere for downstream use.

My code suggestions follow what seemed to be the common choice in the other model native regions.

PR #341, opened 2025-09-22
Add intensity variables

yiyi1991:
This PR adds two sets of intensity variables, emission intensities and energy intensities, for the UNEP FI project. 

How to review:
- Does the scope tag make sense to you? I.e., direct (no suffix), process, indirect, and total. 
- The units are kept as the original energy/emission unit on the energy service unit. Is it better to introduce additional converters so that the value of intensities becomes more readable?
- The building sector is lacking so far due to data availability, as well as the lack of clear definitions in the building sector energy service. Let me know if you actually have suggestions.

### DIFF
--- a/definitions/variable/emissions/emission-intensity.yaml
+++ b/definitions/variable/emissions/emission-intensity.yaml
@@ -1,35 +1,35 @@
-- Emission Intensity|{Level-2 Species}|Industry|Iron and Steel:
-    description: Emission intensity of direct energy-related {Level-2 Species}
+- Emission Intensity|{Level-0 Species}|Industry|Iron and Steel:
+    description: Emission intensity of direct energy-related {Level-0 Species}
       (within Scope 1 and Scope 2) per output from the iron and Steel industry
-    unit: "{Level-2 Species}/Mt"
+    unit: "{Level-0 Species}/Mt"
     tier: 2
-- Emission Intensity|{Level-2 Species}|Industry|Non-Metallic Minerals|{Non-Metallic Minerals Commodity}:
-    description: Emission intensity of direct energy-related {Level-2 Species}
+- Emission Intensity|{Level-0 Species}|Industry|Non-Metallic Minerals|{Non-Metallic Minerals Commodity}:
+    description: Emission intensity of direct energy-related {Level-0 Species}
       (within Scope 1 and Scope 2) per output from the {Non-Metallic Minerals Commodity} industry
-    unit: "{Level-2 Species}/Mt"
+    unit: "{Level-0 Species}/Mt"
     tier: 2
-- Emission Intensity|{Level-2 Species}|Industry|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}:
-    description: Emission intensity of direct energy-related {Level-2 Species}
+- Emission Intensity|{Level-0 Species}|Industry|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}:
+    description: Emission intensity of direct energy-related {Level-0 Species}
       (within Scope 1 and Scope 2) per output from the {Non-Ferrous Metals Commodity} industry
-    unit: "{Level-2 Species}/Mt"
+    unit: "{Level-0 Species}/Mt"
     tier: 2
-- Emission Intensity|{Level-2 Species}|Industry|Chemicals|{Chemicals Commodity}:
-    description: Emission intensity of direct energy-related {Level-2 Species}
+- Emission Intensity|{Level-0 Species}|Industry|Chemicals|{Chemicals Commodity}:
+    description: Emission intensity of direct energy-related {Level-0 Species}
       (within Scope 1 and Scope 2) per output from the {Chemicals Commodity} industry
-    unit: "{Level-2 Species}/Mt"
+    unit: "{Level-0 Species}/Mt"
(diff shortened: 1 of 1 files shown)


definitions/variable/energy/emission-intensity.yaml
```diff
@@ -0,0 +1,107 @@
...
+    description: Emission intensity of direct energy-related CO2 (within Scope 1) per output from the {Chemicals Commodity} industry
+    unit: Mt CO2/Mt
+    tier: 2
+- Emission Intensity|CO2|Industry|Chemicals|{Chemicals Commodity}({scope}):
+    description: Emission intensity of {scope} CO2 per output from the {Chemicals Commodity} industry
+    unit: Mt CO2/Mt
+    tier: 3
+
+- Emission Intensity|CO2|Transportation|Passenger|Road|Light-Duty Vehicle:
+    description: Emission intensity of direct energy-related CO2 (within Scope 1) per output from the passenger transport on roads in light duty vehicles
```
danielhuppmann, line 38: Please replace "output" by the correct metric - I guess passenger-kilometers traveled?

definitions/variable/energy/tag_scope.yaml
```diff
@@ -0,0 +1,8 @@
+- scope:
+    - process:
+        description: the scope of process emissions (within Scope 1)
+    - indirect:
+        description: the scope of indirect emissions (i.e., Scope 2 emissions)
+    - total:
```
danielhuppmann, line 6: The convention in variable names is to not add "Total" as an identifier but have it implicit, like "Emissions|CO2" is the sum (total) of all emissions.

danielhuppmann (changes requested):
Thank you for the PR - see a few comments inline. Please make sure to address these comments throughout the new file (e.g., "output"), I only added comments once for each type of feedback.

definitions/variable/energy/emission-intensity.yaml
danielhuppmann: "energy service demand" is still difficult to understand for non-experts, and the unit is pkm, not energy (service). So why not write "per passenger-distance travelled in light duty vehicles"?

definitions/variable/emissions/emission-intensity.yaml
```diff
@@ -0,0 +1,52 @@
...
+    tier: 2
+- Emission Intensity|Kyoto Gases|Industry|Non-Ferrous Metals|{Non-Ferrous Metals Commodity}:
+    description: Emission intensity of direct energy-related Kyoto Gases (within Scope 1) per output from the {Non-Ferrous Metals Commodity} industry
+    unit: Mt CO2-equiv/Mt
+    tier: 2
+- Emission Intensity|Kyoto Gases|Industry|Chemicals|{Chemicals Commodity}:
+    description: Emission intensity of direct energy-related Kyoto Gases (within Scope 1) per output from the {Chemicals Commodity} industry
+    unit: Mt CO2-equiv/Mt
+    tier: 2
+- Emission Intensity|Kyoto Gases|Transportation|Passenger|Road|Light-Duty Vehicle:
```
lea-hayez, line 41: About aligning to user requirements:

NZBA defined a emissions intensity metrics based on total GHG rather than CO2. Would it make sense aligning this in the description and calculation?
Also, is Scope 1 correct here? I guess we only account for fuel burning in transport?

Page 31: https://www.unepfi.org/wordpress/wp-content/uploads/2024/04/Climate-Target-Setting-for-Automotive-Sector-Financing_Emerging-Practice_NZBA.pdf

definitions/variable/emissions/emission-intensity.yaml
```diff
@@ -0,0 +1,52 @@
...
+    description: Emission intensity of direct energy-related Kyoto Gases (within Scope 1) per passenger-distance travelled on roads in light duty vehicles
+    unit: Mt CO2-equiv/billion pkm
+    tier: 2
+- Emission Intensity|Kyoto Gases|Transportation|Freight|Truck:
+    description: Emission intensity of direct energy-related Kyoto Gases (within Scope 1) per freight-distance travelled on roads in trucks
+    unit: Mt CO2-equiv/billion tkm
+    tier: 2
+- Emission Intensity|CO2|Electricity:
```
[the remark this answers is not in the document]
danielhuppmann: That's a good point, this should be consistent to always have CO2 and GHG indicators.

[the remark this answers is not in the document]
yiyi1991: We do have both (one can find two sets of emission intensities here), 
Emission Intensity|CO2|*
Emissions Intensity|Kyoto Gases|*

but let me try aligning the description with them.

definitions/variable/emissions/emission-intensity.yaml
yiyi1991: Yeah...at the beginning, we were thinking about leaving this as scope 1, and then having "indirect" as scope 2 and "total" as the sum of them. In this way the "total" will then align with them. 

Now we have decided to move the tags to unef-fi workflows. Then let us put scope 1+2 here.

danielhuppmann (approved):
This PR looks ok for me. Computation-methods for the indicators will be added in a follow-up PR.

FYI @IAMconsortium/common-definitions-emissions

PR #352, opened 2025-10-09
Fix description of biomass

danielhuppmann:
When looking at biomass reporting, @volker-krey and I noticed that the secondary-energy tag-list is used both "Secondary Energy" and for "Final Energy" variables. For final energy, biomass should include traditional biomass, but for "production of biomass" as reported in Secondary Energy, traditional biomass is not included.

Hence, the current tag-list description excluding traditional biomass is not correct for Final Energy variables.

This PR simply removes the mention of traditional biomass in the description so that whether traditional biomass is included is implicitly clear from the variable, removing the contradiction.

@IAMconsortium/common-definitions-energy


jkikstra:
LGTM

PR #356, opened 2025-10-17
Create GCAM_8s.yaml

jkikstra:
Copied region definitions over from GCAM 7.4.

PR #357, opened 2025-10-17
Register GCAM 8s

danielhuppmann:
This PR extends #356 by @jkikstra, adding the required mapping file for the region-processing.

@melgeorgev, I simplified the model-registration, assuming that all GCAM 8 versions (so also GCAM 8.1, etc.) will have the same regional resolution). This allows to reduce the number of almost-duplicate files in this repo.

So in this PR, there is a region-hierarchy "GCAM 8", which is then linked in the mapping file to the model "GCAM 8s".

Please let me know whether this makes sense. I will add permissions to submit scenarios for this model once this PR is approved and merged.

### DIFF
--- a/definitions/region/native_regions/GCAM/GCAM_8s.yaml
+++ b/definitions/region/native_regions/GCAM/GCAM_8s.yaml
@@ -1,19 +1,19 @@
-- GCAM 8:
-    - GCAM 8|Africa_Eastern:
+- GCAM 8s:
+    - GCAM 8s|Africa_Eastern:
         countries: [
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Refactored to use model name and hierarchy "GCAM 8s"

melgeorgev:
looks good..thanks

danielhuppmann:
Added permissions for the GCAM modeler team to submit scenarios for "GCAM 8s" to the ssp-submission Scenario Explorer.

PR #360, opened 2025-10-29
adding gcam-ceew-cdr v6

Aravind-Srinivasachari:
We're registering our gcam-ceew-cdr v6 model for the SoCDR study. 

We only model the India region specifically. 

But since gcam is a global model it does have all the other regions. 
Kindly let us know if the yaml file in our model is required to only have the India region.

### DIFF
--- a/mappings/GCAM/GCAM-CEEW-CDR_v6.yaml
+++ b/mappings/GCAM/GCAM-CEEW-CDR_v6.yaml
@@ -1,0 +1,215 @@
+model:
+  - GCAM CEEW CDR v6
+native_regions:
+  - Africa_Eastern: GCAM CEEW CDR v6|Africa_Eastern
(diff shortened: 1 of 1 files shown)


danielhuppmann:
> We only model the India region specifically.

If you only report results for India, it is not necessary to add definitions and mappings.

If you do intend to submit results for all GCAM v6 regions, please confirm that these regions are identical to https://github.com/IAMconsortium/common-definitions/blob/main/definitions/region/native_regions/GCAM/GCAM_6.0.yaml - there‘s an easier way to register your version in this case.

Please advise whether you will report only results for India or all regions.

FYI @CandeBergero

Aravind-Srinivasachari:
Thanks for your response Daniel.

Yes, I can confirm that we only plan to submit results for the India region. Kindly let us know how we can proceed with the model registration in that case. 

Best 
Aravind

PR #361, opened 2025-11-03
Suggesting transport variables and tags

yiyi1991:
This PR proposes new transport sector tags and variables for the NGFS project (Phase VI) but also generally. Key points:

- Add new tag transport mode (`Road`, `Rail`, `Active`, etc.) in addition to the current `vehicle-type` tag (bus, truck, ldv, etc., which only sit in the subcategory of `Road`) and tag `drive-type` (BEV, ICE, etc.), so that mode shift can be presented clearly in the energy service variables. 
- By default, `Aviation` and `Shipping` **do not** include bunkers now.
  - the previous `Domestic Aviation` and `Domestic Shipping` are not precise for activities within model regions, especially in global models
- Two-wheelers are separated from a subcategory of LDV and rescoped as 2W and 3W. 

TODO

- Technical issues still remain in energy service variable namings (being able to match with how they are named in NAVIGATE, how they are defined in SDG indicators)

How to review:

- Read the files changed or generate `nomenclature export-definitions . definitions.xlsx`
- Do the new transport variables (especially mode-tag-related) make sense to you? (defination clear? variables exclusive?)

### DIFF
--- a/definitions/variable/transport/tag_transportation_mode.yaml
+++ b/definitions/variable/transport/tag_transportation_mode.yaml
@@ -1,6 +1,6 @@
 - Transportation Mode:
     - Aviation:
-        description: aviation transport, excluding international bunkers
+        description: aviation transport, excluding activities related to international bunkers
     - Rail:
         description: rail transport
     - Road:
@@ -10,6 +10,6 @@
     - Active:
         description: non-motorized transport
     - Shipping:
-        description: water transport, excluding international bunkers
+        description: water transport, excluding activities related to international bunkers
     - Pipeline:
         description: pipeline transport


yiyi1991:
Dear Rahel, thanks. 

Energy service variables should indeed be included though I would prefer using/improving the tag sets proposed by this PR (e.g., Non-motorized -> Active, no four wheelers, "domestic aviation" does not make sense to global models as "domestic" becomes "within model region" there, Passenger|Bunkers instead of Bunkers|Passenger, etc.). 

Does JUSTMip have an approaching deadline? Feel free to contribute to this PR and pin me to check/review, or kindly consider the tags proposed in this PR and open another to JUSTMip workflow only.

Issue #374, opened 2025-12-12
Add emission aggregation on LULUCF national accounting


RahelMA:
Following Thomas Gasser's approach to incorporate LULUCF national accounting alongside land-use modeling of LULUCF (addressing the Grassi issue) for CO₂ reporting, I'm wondering whether we should also include this approach at the emission aggregation level for policy relevance. 

Specifically, I'm thinking of the variable `Emi|GHG|w/o Bunkers|LULUCF national accounting`, since this is the metric that many countries use as their target in NDC pledges.

@phackstock @tgasser @IoannisDafnomilis

PR #384, opened 2025-12-22
New definitions of emissions in AFOLU sector for comparison with NGHGIs

tgasser:
This is the first of two PRs to initiate a change in the AFOLU-related reporting, following a community-wide consultation carried out within the NewPathways HEU project.

It introduces CO2-specific variables, thereby splitting CO2 and non-CO2 definitions of AFOLU emissions. It also attempts to clean up land-related subcategories, notably fires.

It does not change anything regarding gross emissions variables, as this should be part of a broader effort that includes all sectors.

There's an obvious loss of backward compatibility, although only through renaming as all previously existing subcategories should still be available.

Discussed notably with @flohump, @strefler, but final suggestions my own.

Resulting definitions file attached: [definitions_luc_em_def.xlsx](https://github.com/user-attachments/files/24289632/definitions_luc_em_def.xlsx).

__________

List of changes:

1. Removed `Agricultural Waste Burning` and `Agriculture` subcategories for CO2.
_Rationale_: these should be zero by definition for CO2 because climate-neutral, removing prevents adding possibly non-zero values to anthropogenic emissions.

2. Renamed `Land` subcategory to `LULUCF`, and introduced sub-level categories following IPCC 2006 guidelines' [land types](https://www.ipcc-nggip.iges.or.jp/public/2006gl/vol4.html)) + `Deforestation`. 
_Rationale_: comparison with Grassi et al.'s [NGHGI database](https://essd.copernicus.org/articles/14/4643/2022/) for CO2 requires at least `Forest Land` and `Deforestation`; `HWPs` already present; `Wetlands` important for non-CO2; other subcategories introduced for completeness but can always be ignored.

3. `Fires` were moved two levels up in the tree to `Emissions|{species}|Fires` and removed for CO2.
_Rationale_: only explicit fire management should be included in AFOLU; since it is very unclear whether these are anthropogenic or natural fires we want to force the user to think about it; CO2 removed altogether because either climate-neutral (grassland), already in LULUCF (deforestation fires), or supposed to be in the natural response (other forest or peat fires); non-CO2 may still be necessary for ScenarioMIP (@jkikstra?).

4. Introduced `CO2|AFOLU [NGHGI]|Natural Flux Over Managed Land` as the correction term that allows translation from model/scientific definition to NGHGIs', optionally split over the same land categories as the LULUCF term.
_Rationale_: since this represents only natural processes (response to CO2, climate change, etc.), only a few IAMs can estimate it currently, but it is a necessary component of the total `CO2|AFOLU [NGHGI]` flux.

5. Introduced `CO2|AFOLU [NGHGI]` and `Kyoto Gases|AFOLU [NGHGI]` aggregates that are `CO2|AFOLU` and `Kyoto Gases|AFOLU` + `CO2|AFOLU [NGHGI]|Natural Flux Over Managed Land`.
_Rationale_: for completeness and partially solves [issue 374](https://github.com/IAMconsortium/common-definitions/issues/374).

File structure changes:
- split out the AFOLU sector into a new file (from `emissions` to `emissions-afolu`)
- created a `level-2-species-no-CO2` tag file

Other change:
- corrected level 0 species' units (because used in a previous version, but not used in the end -- thought I'd leave it here anyway)


flohump:
@tgasser — this PR carries `Gross Removals|CO2|AFOLU|Intentional`/`Unintentional`
over unchanged, but I think those two sub-categories should be dropped rather than
relocated: they've been reported only patchily across ScenarioMIP and the
distinction doesn't hold up (reasoning and a State of CDR reference in #224).
Ideally we'd handle this as part of this PR, to avoid merge conflicts on the same
AFOLU block — but I'm happy to open a separate PR if you prefer. Also tagging @strefler and @danielhuppmann here.

strefler:
Agree on removing the intentional / unintentional categories. I can't judge though in which PR this would be placed best.

definitions/variable/emissions/tag_level-0-species.yaml
```diff
@@ -2,8 +2,8 @@
 - Level-0 Species:
     - CO2:
         description: carbon dioxide (CO2)
-        unit: Mt CO2
+        unit: Mt CO2/yr
```
flohump, line 5: This tag is only used for emission intensities, whose unit is built as `{Level-0 Species}/Mt` (emissions per Mt of product, e.g. steel). An intensity has no time dimension, so #366 removed `/yr` from exactly these two lines; this change undoes #366. With `/yr`, all 202 intensity units become e.g. `Mt CO2/yr/Mt` instead of `Mt CO2/Mt`. Suggest reverting (also line 9).

```suggestion
        unit: Mt CO2
```

definitions/variable/emissions/tag_level-0-species.yaml
```diff
@@ -2,8 +2,8 @@
...
     - Kyoto Gases:
         description: Kyoto GHG emissions, including CO2, CH4, N2O and F-gases
         notes: Preferably use AR4 100-year GWPs for aggregation of different gases
-        unit: Mt CO2-equiv
+        unit: Mt CO2-equiv/yr
```
flohump, line 9: Same as line 5.

```suggestion
        unit: Mt CO2-equiv
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+      conversion of non-forest natural land, and drained peatlands
+    unit: "Mt CO2/yr"
+    tier: 1
+    notes: This variable reports the sum of sources and sinks including land-use methods
+      such as soil carbon sequestration, biochar, forestry, and re/afforestation and
+      agroforestry. It uses the model/scientific definition of AFOLU, and does not align
+      with the reporting of national greenhouse gas inventories (NGHGI).
+      For NGHGI-aligned reporting, see 'Emissions|*|AFOLU [NGHGI]'. Emissions reported
+      in this category represent fluxes to/from the atmosphere from/to the land pool.
+      For a breakdown of carbon removals, see the 'Carbon Removal|*' variables.
```
flohump, line 22: `AFOLU [NGHGI]` is defined as `AFOLU` + `Natural Flux Over Managed Land`, so `AFOLU` has to exclude that flux explicitly. Otherwise a model whose carbon densities respond to CO2 and climate counts part of it twice.

```suggestion
    notes: This variable reports the sum of sources and sinks including land-use methods
      such as soil carbon sequestration, biochar, forestry, and re/afforestation and
      agroforestry. It uses the model/scientific definition of AFOLU, and does not align
      with the reporting of national greenhouse gas inventories (NGHGI).
      It excludes the flux reported in 'Emissions|CO2|AFOLU [NGHGI]|Natural Flux Over
      Managed Land', so that the two add up to 'Emissions|CO2|AFOLU [NGHGI]'.
      For NGHGI-aligned reporting, see 'Emissions|*|AFOLU [NGHGI]'. Emissions reported
      in this category represent fluxes to/from the atmosphere from/to the land pool.
      For a breakdown of carbon removals, see the 'Carbon Removal|*' variables.
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+    tier: 3
+
+
+# Subcategories including CO2
+- Emissions|{Level-2 Species}|AFOLU|LULUCF:
+    description: Emissions of {Level-2 Species} from forestry and other land use and
+      land use change. Removals in this variable include agroforestry, re/afforestation,
+      biochar, soil carbon management and forest management.
+    unit: "{Level-2 Species}"
+    tier: 1
```
flohump, line 125: Declaring the eight subcategories as components makes "they sum to LULUCF" part of the definition, as for the other aggregates in this file.

```suggestion
    tier: 1
    components:
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Deforestation
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Forest Land
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Cropland
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Grassland
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Settlements
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Other Land
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Wetlands
      - Emissions|{Level-2 Species}|AFOLU|LULUCF|Harvested Wood Products
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+- Emissions|{Level-2 Species}|AFOLU|LULUCF|Deforestation:
+    description: Emissions and removals from any land converted from forest to any other land,
+      except wetlands (IPCC 2006 categories 3B2bi, 3B3bi, 3B5bi, 3B6bi).
+    unit: "{Level-2 Species}"
+    tier: 2
+    notes: All lands converted from Forest should go here, excluding wetlands-related transitions. 
+      This also excludes Harvest Wood Products.
```
flohump, line 132: Typo.

```suggestion
      This also excludes Harvested Wood Products.
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+- Emissions|{Level-2 Species}|AFOLU|LULUCF|Wetlands:
+    description: Emissions from managed wetlands (e.g., peatland),
+      including emissions from drained wetlands used for agriculture, forestry or other uses
+      (IPCC 2006 categories 3B1biii, 3B2biii and 3B3biii, 3B5biv, 3B6biv),
+      and emissions from rewetted wetlands (IPCC 2006 category 3B4)
+    unit: "{Level-2 Species}"
+    tier: 3
+    notes: Includes reservoirs as a managed sub-division and
+      natural rivers and lakes as unmanaged sub-divisions. Includes drained
+      and rewetted wetlands, does not include natural emissions from intact wetlands.
```
flohump, line 167: With CO2 `Fires` removed, CO2 from fires on drained peatland needs a home, or it drops out of AFOLU; the GCB counts peat burning and peat drainage in ELUC (Friedlingstein et al. 2025, https://doi.org/10.5194/essd-17-965-2025). IPCC 2006 books drained organic soils within each land category, so the note also says where they go here.

```suggestion
    notes: Includes reservoirs as a managed sub-division and
      natural rivers and lakes as unmanaged sub-divisions. Includes drained
      and rewetted wetlands, does not include natural emissions from intact wetlands.
      Also includes drained organic soils (peatland) under any land use, which are
      therefore not reported in the other land categories, and CO2 from fires on
      drained peatland.
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+- Emissions|{Level-2 Species}|AFOLU|LULUCF|Harvested Wood Products:
+    description: Emissions and removals from harvested wood products (HWP)
+      (no clear IPCC 1996 category - IPCC 2006 category 3D1)
+    unit: "{Level-2 Species}"
+    tier: 3
```
flohump, line 172: Current submissions report HWP with opposite signs, so the convention should be explicit.

```suggestion
    tier: 3
    notes: Net flux of the product pool. Carbon entering the pool is reported negative,
      emissions from decay or burning of products positive. The harvest itself is reported
      as an emission in the land category the wood comes from.
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+- Emissions|{Level-2 Species}|AFOLU|Other:
+    description: Emissions and removals in AFOLU sector that cannot be accommodated in other
+      categories (no clear IPCC 1996 category - IPCC 2006 category 3D2)
+    unit: "{Level-2 Species}"
+    tier: 3
```
flohump, line 177: CO2 from agriculture is not zero by definition: liming and urea application (IPCC 2006 3C2, 3C3) are CO2 sources within the old `Agriculture` definition (3A and 3C except 3C1). With CO2 `Agriculture` dropped, they need a home.

```suggestion
    tier: 3
    notes: For CO2, this includes emissions from liming and urea application
      (IPCC 2006 categories 3C2 and 3C3).
```

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+    description: Emissions from agricultural waste burning
+      (IPCC 1996 category 4F - partially IPCC 2006 category 3C1)
+    unit: "{Level-2 Species w/o CO2}"
+    tier: 2
+    notes: Also sometimes known as 'biomass burning' or 'open agricultural burning'
+
+
+# Fire Emissions only for Non-CO2
+# /!\ Moved out of AFOLU because current definition is unclear.
+- Emissions|{Level-2 Species w/o CO2}|Fires:
```
flohump, line 223: `Emissions|BC`, `Emissions|CH4` etc. are defined as the sum of AFOLU, Energy, Industrial Processes, Waste, Other Capture and Removal, Product Use and Other (`emissions.yaml`, lines 14-21). Fires used to be part of AFOLU; as `Emissions|{species}|Fires` they are in none of these, so the totals no longer include fire emissions. Either add `Fires` to that list or keep them as `AFOLU|Fires`. @jkikstra, does anything downstream (emissions harmonization, gridding, climate emulators) rely on fire emissions being part of AFOLU or of the species totals? Which option would you prefer?

definitions/variable/emissions/emissions-afolu.yaml
```diff
@@ -0,0 +1,280 @@
...
+    notes: Removals reported in this category represent fluxes from the atmosphere to the land pool.
+- Gross Removals|CO2|AFOLU|Unintentional:
+    description: Gross removals of carbon dioxide (CO2) from agriculture, forestry
+      and other land use (IPCC 1996 category 4, 5 - IPCC 2006 category 3),
+      due to unintentional or indirect removals,
+      e.g. from regrowth of natural vegetation after land abandonment, regrowth of forests
+      after wood harvest or regrowth of timber plantations
+    unit: Mt CO2/yr
+    tier: 2
+    notes: Removals reported in this category represent fluxes from the atmosphere to the land pool.
```
flohump, line 280: As in my comment of 26 Jun (agreed by @strefler): I suggest dropping the Intentional/Unintentional split, including in the description. Submissions report it patchily and with opposite signs; reasoning in #224. The rewrite also avoids "indirect removals", as indirect effects now belong to `Natural Flux Over Managed Land`.

```suggestion
    description: Gross removals of carbon dioxide (CO2) from agriculture, forestry
      and other land use (IPCC 1996 category 4, 5 - IPCC 2006 category 3),
      e.g. from re/afforestation, regrowth of natural vegetation after land abandonment,
      regrowth of forests after wood harvest, regrowth of timber plantations
      and soil carbon sequestration
    unit: Mt CO2/yr
    tier: 1
    notes: Removals reported in this category represent fluxes from the atmosphere to the land pool.
```

flohump (changes requested):
I took the time to go through this once more in detail, with help from Claude AI.

I support the new structure; it matches what we discussed. The CO2 AFOLU subcategories that this PR removes or moves are used inconsistently in the ScenarioMIP submissions: `Emissions|CO2|AFOLU|Agriculture`, `...|Agricultural Waste Burning`, `...|Land|Fires` and `...|Land|Other` are non-zero in only one or two models each, and `...|Land|Land Use and Land-Use Change` covers a different part of AFOLU in each model. So it is fine to remove them.

I added 9 inline suggestions that can be committed directly, plus one comment on fires. Two points block merging:

1. The Level-0 unit change turns 202 emission-intensity units into `.../yr/...`.
2. Fire emissions are no longer part of the `Emissions|{species}` totals (@jkikstra).

I also suggest dropping Intentional/Unintentional, as agreed with @strefler. #417 edits the same lines, so one of the two needs a rebase (@danielhuppmann).

With all suggestions applied, the project validation (`nomenclature validate-project`) and the tests pass.

PR #398, opened 2026-03-02
add new transport subsectors to tag_all_sectors.yaml

robertpietzcker:
- added Transportation|Road: To be able to compare with real-world sources (eg IEA), the total for road transport is needed. 
- added Transportation|Rail|Passenger / Transportation|Rail|Freight: Rail is the only transportation sector that provides both passenger and freight energy services and is in many models modeled separately - so it is important to have this differentiation also in the reporting.

### DIFF
--- a/definitions/variable/energy/tag_all_sectors.yaml
+++ b/definitions/variable/energy/tag_all_sectors.yaml
@@ -129,8 +129,17 @@
     - Transportation|Truck:
         description: trucking sector
         tier: ^1
+    - Transportation|Road:
+        description: road transportation sector - both passenger and freight
+        tier: ^1
     - Transportation|Rail:
         description: rail transportation sector
+        tier: ^1
+    - Transportation|Rail|Passenger:
+        description: passenger rail transportation sector
+        tier: ^1
+    - Transportation|Rail|Freight:
+        description: freight rail transportation sector
         tier: ^1
     - Transportation|Domestic Aviation:
         description: domestic aviation sector


danielhuppmann:
Pinging @IAMconsortium/common-definitions-transport @IAMconsortium/common-definitions-coordination for comment

I am wondering if we should also add a sub-division between passenger vs. freight on roads...

robertpietzcker:
> I am wondering if we should also add a sub-division between passenger vs. freight on roads...

One could clearly add that, but it also would be possible to calculate directly from the LDV/Buses/Trucks reporting. 
For rail, these separate categories did not exist until now afaik, so it is crucial to add them. 

(not saying I am against adding the road split, I just wanted to explain why I think the one is necessary and the other is optional)

khaeru:
iTEM uses the following concepts:
- SERVICE: P (Passenger), F (Freight), _T (Total).
- MODE: include RAIL, ROAD, others, _T.
- VEHICLE_TYPE: many, e.g. BUS, LDV, TRUCK, others, _T.

In the full product of all these code lists, some combinations are meaningless and will never be used, for instance:
- (SERVICE=F, MODE=ROAD, VT=BUS) —buses are not used for freight service.
- (SERVICE=P, MODE=RAIL, VT=BUS) —buses do not operate on rails.

It seems like "sector" here is being used to refer to combinations of (SERVICE, MODE).

> I am wondering if we should also add a sub-division between passenger vs. freight on roads...
>
> > it also would be possible to calculate directly from the LDV/Buses/Trucks reporting.

This presumes that there are 1 or more VEHICLE_TYPEs that collectively add up to either (SERVICE=F, MODE=ROAD, VT=_T) or (SERVICE=P, MODE=ROAD, VT=_T) (put another way: that the available data covers either _all_ freight or _all_ passenger road vehicle types), such that one can be subtracted from (SERVICE=_T, MODE=ROAD, VT=_T) to get the other. 

Maybe that's the case in actual collections of model data—I don't know. If not, then having (SERVICE=F, MODE=ROAD, VT=_T) etc. explicitly would probably help.

danielhuppmann:
I think it would be best to be pragmatic and consider that some models may not have the same (or any) level of detail of the types of vehicles, so that would speak for having both
```
- Transportation|Road|{Type of Vehicle}
```
and 
```
- Transportation|Road|Passenger
- Transportation|Road|Freight 
```
listed explicitly.

danielhuppmann:
If there are no objections, I will merge this PR tomorrow and then do a follow-up PR to add the road freight/passenger subcategories.

PR #399, opened 2026-03-24
Add passenger vs. freight categories for road transport

danielhuppmann:
This PR adds the freight vs. road categories for road transportation as a direct follow-up to #398. It also harmonizes some descriptions for consistency


danielhuppmann:
FYI @IAMconsortium/common-definitions-coordination

PR #401, opened 2026-03-25
feat: Add WITCH 6.0-EU28 config files

lolow:
- Add a new configuration file for WITCH 6.0-EU28 defining regional country groupings.
- Add new WITCH 6.0-EU28 model configuration with regional mappings and classifications.


danielhuppmann (approved):
Looks good, thank you!

As an idea for a follow-up PR, you could also add mappings for G20 members to the WITCH mapping files, so that it is easier to compare the results for large economies in MIPs, see this example https://github.com/IAMconsortium/common-definitions/blob/c00367f7c65e4453cad48ccca7e994f059d236d5/mappings/REMIND-MAgPIE/REMIND-MAgPIE_3.6-4.13.yaml#L99

PR #410, opened 2026-06-01
Added IF v8.54

VassilisDaioglou:
Added region definition for IFs v 8.54


VassilisDaioglou:
@danielhuppmann Hi Daniel, im trying to inset the region mapping for the IFs model but im getting a validation error i dont understand. Help!

mappings/IFs/IFs_v8.54.yaml
```diff
@@ -0,0 +1,889 @@
+model:
+  -IFs 8.54
+
+native regions:
```
danielhuppmann:
```suggestion
native_regions:
```

mappings/IFs/IFs_v8.54.yaml
```diff
@@ -0,0 +1,889 @@
...
+    - IDN
+    - ITA
+    - JPN
+    - KOR
+    - MEX
+    - RUS
+    - SAU
+    - ZAF
+    - TUR
+    - USA
```
danielhuppmann: There is no aggregate-G20-region defined (yet), so this isn't going to work. Do you think that it would be very relevant?
```suggestion
```

danielhuppmann (commented):
There is a whitespace instead of an underscore, so the yaml file couldn't be parsed correctly.

About the region-definitions themselves, it looks to me like this model has all the countries of the world, so you could simply upload the country names as is, right?

In which project do you want to use this model?

VassilisDaioglou:
@danielhuppmann  Thanks for your changes! This is for the Sustainability-MIP, part of PRISMA. 

All checks pass now, can this be merged?

danielhuppmann:
@dc-almeida @lisahligono, I think that we do not yet have a **CountryProcessor** - I guess that it would make sense to add one to **nomenclature** similar to the [NutsProcessor](https://github.com/IAMconsortium/nomenclature/blob/ad5fa1dde1fa90863c309ece7c1d38a92da8cf7d/nomenclature/processor/nuts.py#L64). What do you think?

danielhuppmann:
> All checks pass now, can this be merged?

@VassilisDaioglou , because this model uses only countries, there is no need for the model prefix in the region names - that just adds wight for little gain. So I implemented this mapping directly in the prisma-workflow repo, together with some observations about possible inconsistencies.

Closing in favour of https://github.com/iiasa/prisma-workflow/pull/23

PR #412, opened 2026-06-16
Register C3IAM model

ceepliulijing:
This PR registers C3IAM_v3.0 by adding the model native region definitions and model mapping files following the nomenclature model-registration instructions.

Reviewers: @danielhuppmann @phackstock


definitions/region/native_regions/C3IAM_v3.0.yaml
```diff
@@ -0,0 +1,73 @@
...
+        countries: [
+          China, Hong Kong, Taiwan
+        ]
+    - C3IAM 3.0|Japan:
+        countries: Japan
+    - C3IAM 3.0|Russia:
+        countries: Russian Federation
+    - C3IAM 3.0|India:
+        countries: India
+    - C3IAM 3.0|OBU:
```
danielhuppmann: It would be more intuitive to users to rename the region so that it doesn't use abbreviations
```suggestion
    - C3IAM 3.0|Canada and Oceania:
```

mappings/C3IAM_v3.0.yaml
```diff
@@ -0,0 +1,99 @@
+model:
+  - C3IAM 3.0
+
+native_regions:
+  - USA: C3IAM 3.0|USA
+  - China: C3IAM 3.0|China
+  - Japan: C3IAM 3.0|Japan
+  - Russia: C3IAM 3.0|Russia
+  - India: C3IAM 3.0|India
+  - OBU: C3IAM 3.0|OBU
```
danielhuppmann:
```suggestion
  - OBU: C3IAM 3.0|Canada and Oceania
```

danielhuppmann (commented):
This looks ok, thank you - but you may want to consider replacing the non-obvious abbreviation with longer, more descriptive names. See an example suggestion below.

ceepliulijing:
Thanks a lot for the suggestion. We have added comments to clarify the abbreviations (e.g., OBU = Other Branches of Umbrella Group, OWE = Other Western Europe). We would like to keep the existing region names because they have been used consistently in our previous published studies, which helps maintain consistency across related analyses.

danielhuppmann (approved):
Looks good, thank you!

PR #414, opened 2026-06-24
Register GRACE_v26

taoyuancicero:
[@danielhuppmann](https://github.com/danielhuppmann) and [@phackstock](https://github.com/phackstock)

GRACE_v26 will participate in the IAMC diagnostics exercise, and we are encouraged to register the model here.

### DIFF
--- a/definitions/region/GRACE_v26.yaml
+++ b/definitions/region/GRACE_v26.yaml
@@ -1,0 +1,272 @@
+- GRACE v26:
+  - GRACE v26|Oceania:
+      countries:
+      - American Samoa
+      - Australia
+      - Fiji
+      - French Polynesia
+      - Guam
+      - Kiribati
+      - Marshall Islands
+      - Micronesia
+      - Nauru
+      - New Caledonia
+      - New Zealand
+      - Northern Mariana Islands
+      - Palau
+      - Papua New Guinea
+      - Samoa
+      - Solomon Islands
+      - Tonga
+      - Tuvalu
+      - Vanuatu
+  - GRACE v26|East Asia:
+      countries:
+      - China
+      - Hong Kong
+      - Japan
+      - Macao
+      - Mongolia
+      - North Korea
+      - South Korea
+      - Taiwan
+  - GRACE v26|South-East Asia:
+      countries:
+      - Brunei Darussalam
+      - Cambodia
+      - Indonesia
+      - Laos
+      - Malaysia
+      - Myanmar
+      - Philippines
+      - Singapore
+      - Thailand
+      - Timor-Leste
+      - Viet Nam
+  - GRACE v26|South Asia:
+      countries:
+      - Afghanistan
+      - Bangladesh
+      - Bhutan
+      - India
+      - Maldives
+      - Nepal
+      - Pakistan
+      - Sri Lanka
+  - GRACE v26|West Asia:
+      countries:
+      - Bahrain
+      - Cyprus
+      - Iran
+      - Iraq
+      - Israel
+      - Jordan
+      - Kuwait
+      - Lebanon
+      - Oman
+      - Palestine
+      - Qatar
+      - Saudi Arabia
+      - Syria
+      - Turkey
+      - United Arab Emirates
+      - Yemen
+  - GRACE v26|North America:
+      countries:
+      - Bermuda
+      - Canada
+      - Greenland
+      - Mexico
+      - United States
+  - GRACE v26|South America:
+      countries:
+      - Anguilla
+      - Antigua and Barbuda
+      - Argentina
+      - Belize
+      - Bolivia
+      - Bonaire, Sint Eustatius and Saba
+      - Brazil
+      - Chile
+      - Colombia
+      - Costa Rica
+      - Cuba
+      - Curaçao
+      - Dominica
+      - Dominican Republic
+      - Ecuador
+      - El Salvador
+      - Guatemala
+      - Guyana
+      - Haiti
+      - Honduras
+      - Jamaica
+      - Martinique
+      - Montserrat
+      - Nicaragua
+      - Panama
+      - Paraguay
+      - Peru
+      - Puerto Rico
+      - Suriname
+      - Trinidad and Tobago
+      - Uruguay
+      - Venezuela
+  - GRACE v26|Europe:
+      countries:
+      - Åland Islands
+      - Albania
+      - Austria
+      - Belgium
+      - Bulgaria
+      - Croatia
+      - Czechia
+      - Denmark
+      - Estonia
+      - Finland
+      - France
+      - Germany
+      - Greece
+      - Hungary
+      - Iceland
+      - Ireland
+      - Italy
+      - Latvia
+      - Liechtenstein
+      - Lithuania
+      - Luxembourg
+      - Malta
+      - Netherlands
+      - Norway
+      - Poland
+      - Portugal
+      - Romania
+      - Serbia
+      - Slovakia
+      - Slovenia
+      - Spain
+      - Sweden
+      - Switzerland
+      - United Kingdom
+  - GRACE v26|Africa:
+      countries:
+      - Algeria
+      - Angola
+      - Benin
+      - Botswana
+      - Burkina Faso
+      - Burundi
+      - Cabo Verde
+      - Cameroon
+      - Central African Republic
+      - Chad
+      - Comoros
+      - Congo
+      - Côte d'Ivoire
+      - Democratic Republic of the Congo
+      - Djibouti
+      - Egypt
+      - Equatorial Guinea
+      - Eritrea
+      - Eswatini
+      - Ethiopia
+      - Falkland Islands (Malvinas)
+      - French Guiana
+      - Gabon
+      - Gambia
+      - Ghana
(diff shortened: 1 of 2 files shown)


definitions/region/GRACE_v26.yaml
```diff
@@ -0,0 +1,210 @@
...
+      countries:
+      - China
+      - Hong Kong
+      - Japan
+      - Macao
+      - Mongolia
+      - North Korea
+      - South Korea
+      - Taiwan
+  - South-east Asia:
```
danielhuppmann: Capitalise always...
```suggestion
  - GRACE v26|South-East Asia:
```

definitions/region/GRACE_v26.yaml
```diff
@@ -0,0 +1,210 @@
+- GRACE v26:
+  - Oceania:
```
danielhuppmann: Native regions should have a prefix with the model-version-hierarchy.
```suggestion
  - GRACE v26|Oceania:
```

mappings/GRACE_v26.yaml
```diff
@@ -0,0 +1,31 @@
+model: GRACE v26
+native_regions:
+- OCE: Oceania
```
danielhuppmann: Rename-targets must also include the prefix.
```suggestion
- OCE: GRACE v26|Oceania
```

mappings/GRACE_v26.yaml
```diff
@@ -0,0 +1,31 @@
...
+- SEAS: South-east Asia
+- SAS: South Asia
+- WAS: West Asia
+- NAMR: North America
+- SAMR: South America
+- EU: European Union
+- AFR: Africa
+- RoW: Rest of the World
+common_regions:
+- Latin America (R5):
```
danielhuppmann, line 36: This mapping is not complete, please add mappings for all R5 and R10 - and maybe also R9 regions?

definitions/region/GRACE_v26.yaml
```diff
@@ -0,0 +1,210 @@
...
+      - Nicaragua
+      - Panama
+      - Paraguay
+      - Peru
+      - Puerto Rico
+      - Suriname
+      - Trinidad and Tobago
+      - Uruguay
+      - Venezuela
+  - European Union:
```
danielhuppmann: Note that this isn't the (current) EU members, so better to rename to avoid confusion.
```suggestion
  - GRACE v26|Europe:
```

mappings/GRACE_v26.yaml
```diff
@@ -0,0 +1,31 @@
+model: GRACE v26
```
danielhuppmann, line 1: If you intend to submit scenarios as model-version GRACE v26, that's ok  - but you can also use one region-hierarchy "GRACE v26" for multiple minor/patch-versions of the model (assuming that native-regions will not change between major versions)
```suggestion
model: GRACE v26.0.0
```

mappings/GRACE_v26.yaml
```diff
@@ -0,0 +1,31 @@
...
+native_regions:
+- OCE: Oceania
+- EAS: East Asia
+- SEAS: South-east Asia
+- SAS: South Asia
+- WAS: West Asia
+- NAMR: North America
+- SAMR: South America
+- EU: European Union
```
danielhuppmann:
```suggestion
- EU: GRACE v26|European Union
```

danielhuppmann (changes requested):
Thank you, see suggestions and comments below.

mappings/GRACE_v26.yaml
taoyuancicero: Other native regions cannot exactly map to R5, R10, and R9. For example, we do not have a separate region like China, Japan, and OECD. How could I do the mapping?

mappings/GRACE_v26.yaml
taoyuancicero: We are suggested to use only v26, since we will not change the regions for minor versions in this exercise.

mappings/GRACE_v26.yaml
taoyuancicero: Will modify this and other issues. Should I then close the conversation and send a new pull request?

mappings/GRACE_v26.yaml
danielhuppmann: `OECD & EU (R5)` seems close enough to NAMR and EU, and `Asia (R5)` is close to SAS, SEAS and EAS - except for Japan, but there can often not be perfect mappings.

Not sure where Russia is in your region-resolution (maybe RoW) - that would be a good guess for `Reforming Economies (R5)`?

The rule-of-thumb is that you should have all native-regions mapped to either one of the R5 regions or to `Other (R5)` so that the sum of the R5 regions (including Other") is equal to global results.

danielhuppmann, line 36: I overlooked that aggregation from native-regions to global values should also be added explicitly (unless you want to include "World" in your scenario submission files explicitly?

```suggestion
- World:
  - OCE
  - EAS
  - SEAS
  - SAS
  - WAS
  - NAMR
  - SAMR
  - EU
  - AFR
  - RoW
- Latin America (R5):
```

mappings/GRACE_v26.yaml
danielhuppmann: You can simply push to your `main` branch and this current Pull Request will be automatically updated.

mappings/GRACE_v26.yaml
danielhuppmann: Sorry, I wasn't clear - which model version will you use for submitting to the IAMC-diagnistics exercise "GRACE v26" or "GRACE v26.0" or something else?

definitions/region/GRACE_v26.yaml
```diff
@@ -0,0 +1,210 @@
...
+      - South Sudan
+      - Sudan
+      - Tanzania
+      - Togo
+      - Tunisia
+      - Uganda
+      - Western Sahara
+      - Zambia
+      - Zimbabwe
+  - GRACE v26|Rest of the World
```
danielhuppmann: Please add the countries-attribute here as well, so that (at least) Russia is listed in the native-region-definition file.

mappings/GRACE_v26.yaml
taoyuancicero: GRACE v26

mappings/GRACE_v26.yaml
```diff
@@ -0,0 +1,75 @@
...
+  - SAMR
+- Middle East & Africa (R9):
+  - WAS
+  - AFR
+- Other (R9):
+  - OCE
+  - EAS
+  - SEAS
+  - SAS
+  - NAMR
```
danielhuppmann: If only four of 9 regions can be mapped explicitly, better to remove the R9-mapping altogether to avoid confusion. Sorry for suggesting it earlier.
```suggestion
```

mappings/GRACE_v26.yaml
danielhuppmann: Ok, thanks, clear now.

definitions/region/GRACE_v26.yaml
```diff
@@ -0,0 +1,271 @@
...
+      - Sudan
+      - Tanzania
+      - Togo
+      - Tunisia
+      - Uganda
+      - Western Sahara
+      - Zambia
+      - Zimbabwe
+  - GRACE v26|Rest of the World:
+      - Andorra
```
danielhuppmann, line 212: `countries` attribute is missing here:

taoyuancicero: The countries are based on the Excel template file at https://raw.githubusercontent.com/IAMconsortium/nomenclature/main/templates/model-registration-template.xlsx. I could not see why some countries are still missing here.

danielhuppmann: Not an individual country but the word "countries"...
```suggestion
      countries:
     - Andorra
```

definitions/region/GRACE_v26.yaml
```diff
@@ -0,0 +1,272 @@
...
+      - Tanzania
+      - Togo
+      - Tunisia
+      - Uganda
+      - Western Sahara
+      - Zambia
+      - Zimbabwe
+  - GRACE v26|Rest of the World:
+      countries:
+     - Andorra
```
danielhuppmann: One more whitespace, then the tests should pass
```suggestion
      - Andorra
```

mappings/GRACE_v26.yaml
```diff
@@ -0,0 +1,59 @@
...
+- North America (R10):
+  - NAMR
+- Africa (R10):
+  - AFR
+- Reforming Economies (R10):
+  - RoW
+- Other (R10):
+  - EAS
+  - SEAS
+  - SAS
```
danielhuppmann: One more suggestion, it seems that your regions for Asia actually map quite well to the three Asian regions of the R10 aggregation
```suggestion
- South Asia (R10):
  - SAS
- China+ (R10):   
  - EAS
- Rest of Asia (R10):   
  - SEAS
```

danielhuppmann (approved):
Looks good, sorry for the many iterations!

PR #415, opened 2026-06-24
Adding MESSAGEix-GLOBIOM-GAINS 2.1-BMT-R12 model

volker-krey:
Note that native region names use 'MESSAGEix-GLOBIOM-GAINS 2.1-R12' as a prefix similar to the 'MESSAGEix-GLOBIOM-GAINS 2.1-M-R12' model. So, both don't include the '-M' and '-BMT' insert in the region name and therefore share the same native region names which is a benefit when comparing scenarios across the two model variants.


danielhuppmann:
Thanks, but it is not necessary to duplicate the mapping, you can simply add "MESSAGEix-GLOBIOM-GAINS 2.1-BMT-R12" to https://github.com/IAMconsortium/common-definitions/blob/main/mappings/MESSAGEix-GLOBIOM/MESSAGEix-GLOBIOM_2.1-R12.yaml

Closing this PR, please reopen if the mapping is different between the `BMT` model and the default and `M` versions.

PR #416, opened 2026-06-24
Adding MESSAGEix-GLOBIOM-GAINS 2.1-BMT-R12 model

volker-krey:
Adding MESSAGEix-GLOBIOM-GAINS 2.1-BMT-R12 model with identical native region mapping to that of MESSAGEix-GLOBIOM-GAINS 2.1-M-R12.


danielhuppmann (approved):
Thank you!

Note that mappings have to be imported also to the project-specific workflow repo.

PR #417, opened 2026-06-24
Validate emissions with lower bound

### LINKED ISSUE #339: Introduce `lower_bound` on non-CO2 emissions
Would it be a good idea to implement a lower bound of zero for all species where no net-negative/removal technologies have been invented yet? 
This could ensure that reporting issues would be captured sooner.

danielhuppmann:
This PR adds validation of non-negativity (lower bound = 0) to all emissions species except CO2 and Kyoto GHG, and enforces non-negative values for gross emission and gross removals of CO2.

FYI @jkikstra @IAMconsortium/common-definitions-emissions @phackstock @dc-almeida 

Closes #339

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -100,95 +100,11 @@
     lower_bound: 0
     notes: Removals reported in this category represent fluxes from the atmosphere to the land pool.
 
-- Emissions|{Level-3 Species}|AFOLU|Agriculture:
-    description: Emissions of {Level-3 Species} from the agriculture sector
-      (IPCC 1996 category 4 - IPCC 2006 category 3A and 3C except 3C1)
+- Emissions|{Level-3 Species}|AFOLU|{AFOLU Emissions}:
+    description: Emissions of {Level-3 Species} from {AFOLU Emissions}
     unit: "{Level-3 Species}"
     tier: 2
-- Emissions|{Level-3 Species}|AFOLU|Agriculture|Livestock:
-    description: Emissions of {Level-3 Species} from livestock in the agriculture sector
-      (IPCC 1996 category 4A and 4B - IPCC 2006 category 3A)
-    unit: "{Level-3 Species}"
-    tier: 2
-- Emissions|{Level-3 Species}|AFOLU|Agriculture|Livestock|Enteric Fermentation:
-    description: Emissions of {Level-3 Species} from enteric fermentation of livestock
-      (IPCC 1996 category 4A - IPCC 2006 category 3A1)
-      in the agriculture sector
-    unit: "{Level-3 Species}"
-    tier: 3
-- Emissions|{Level-3 Species}|AFOLU|Agriculture|Livestock|Manure Management:
-    description: Emissions of {Level-3 Species} from processes involving
-      manure (waste) management in Livestock in the agriculture sector
-      (IPCC 1996 category 4B - IPCC 2006 category 3A2)
-    unit: "{Level-3 Species}"
-    tier: 3
-- Emissions|{Level-3 Species}|AFOLU|Agriculture|Rice:
-    description: Emissions of {Level-3 Species} from rice cultivation
-      (IPCC 1996 category 4C - IPCC 2006 category 3C7)
-    unit: "{Level-3 Species}"
-    tier: 3
-- Emissions|{Level-3 Species}|AFOLU|Agriculture|Managed Soils:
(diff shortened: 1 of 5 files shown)


definitions/variable/emissions/emissions.yaml
```diff
@@ -25,14 +24,16 @@
...
+    notes: Net emissions 'Emissions|CO2' = 'Gross Emissions|CO2' - 'Gross Removals|CO2'.
 - Gross Removals|CO2:
     description: Gross removals of carbon dioxide (CO2) excluding positive emissions,
       including removals from bioenergy with CCS (BECCS), direct air carbon capture
       and agriculture, forestry and other land use (AFOLU)
     unit: Mt CO2/yr
     tier: 1
-    notes: Net emissions 'Emissions|CO2' = 'Gross Emissions|CO2' + 'Gross Removals|CO2'.
+    lower_bound: 0
+    notes: Net emissions 'Emissions|CO2' = 'Gross Emissions|CO2' - 'Gross Removals|CO2'.
```
phackstock, line 36: The notes should be different to the ones for `Gross Emissions|CO2`, no?

phackstock (approved):
Looks good to me, one small question in line. 
Did you check that the replacement of the `lower_bound` values works and a number is actually read instead of a string? Does the `lower_bound` replacement also work for the `None` values and is the validator (or lack thereof in this case) then applied correctly?

danielhuppmann:
> Did you check that the replacement of the `lower_bound` values works and a number is actually read instead of a string? 

Yes

> Does the `lower_bound` replacement also work for the None values and is the validator (or lack thereof in this case) then applied correctly?

Yes. (It has to be **None** in the yaml-sense of no value, not the string "None")

definitions/variable/emissions/emissions.yaml
danielhuppmann: Don't understand what you mean, the equation "Net Emissions = Gross Emissions - Gross Removals" should always be true. And I think it's less confusing (and less error-prone) to always use exactly the same ordering of terms.

flohump:
In general I agree with the sign change — the +/− inconsistency should be fixed.
One concern before it lands:

@danielhuppmann — could we check the ScenarioMIP impact before merging? The
ScenarioMIP submission workflow (`iiasa/ssp-submission-workflow`) picks up changes
from this repo's `main` automatically — there's no fixed version. So merging the
sign change would immediately change validation for submissions already in: most
models reported `Gross Removals|CO2|AFOLU` as negative and would suddenly fail the
new `lower_bound: 0` check. Could we freeze the workflow to a fixed version before
this merges, and update it once ScenarioMIP submissions are final?
Also tagging @laurinks here

(There's a related structural question — whether the `Intentional`/`Unintentional`
split should stay at all — but that belongs in #224, where I've commented.)

gidden:
One additional note - in work we did with @tgasser, a small number of assessed scenarios had negative values for carbon removal on land due to the flux definitions applied.

In principle, I agree this is counter intuitive and it may be best to just have a check that all gross removals are strictly positive. But based on how land carbon flux is defined, it is possible (but unlikely) to provide a valid negative flux for this variable.

jkikstra:
> ScenarioMIP submission workflow (iiasa/ssp-submission-workflow) ... Could we freeze the workflow to a fixed version before this merges, and update it once ScenarioMIP submissions are final?

@phackstock would you be able to take care of that?

tgasser:
> One additional note - in work we did with @tgasser, a small number of assessed scenarios had negative values for carbon removal on land due to the flux definitions applied.
> 
> In principle, I agree this is counter intuitive and it may be best to just have a check that all gross removals are strictly positive. But based on how land carbon flux is defined, it is possible (but unlikely) to provide a valid negative flux for this variable.

Just want to insist: for the land sector, in some cases, it is possible to get a negative removal, i.e. a flux to the atmosphere. I'd be careful with the validation, here.

phackstock:
> > ScenarioMIP submission workflow (iiasa/ssp-submission-workflow) ... Could we freeze the workflow to a fixed version before this merges, and update it once ScenarioMIP submissions are final?
> 
> @phackstock would you be able to take care of that?

~~@jkikstra, the validation is not automatically applied anyway, so merging this would not cause any uploads to fail without explicitly setting the validator in the workflow. This should be fine to merge without having to pin common-definitions in the ssp-submission-workflow.~~

Never mind, as @danielhuppmann pointed out, the validation would be automatically applied, we'll apply the pin once this PR gets merged to avoid any issues.

PR #423, opened 2026-07-24
Register medeas spain v1.0.0

earamendia:
Hi there
Attempting to register a new model, I've tried to follow closely instructions but first time I do this so not totally sure it is all OK
I did run "nomenclature validate-project /project/folder/" though and didn't get any error
Thanks


danielhuppmann:
Thank you @earamendia for your effort to register a new model in **common-definitions**. However, in cases where models only represent a single country, it is much easier, no need to do any registration wth definitions and mappings.

Please simply send me an email with the project for which you want to submit scenarios and your user name at https://manager.ece.iiasa.ac.at 

Closing this PR because it is not necessary.

PR #428, opened 2026-08-24
Register GCAM-PKU 7.0

FanWu1020:
Hi @danielhuppmann @phackstock,

This PR registers GCAM v7.0 (GCAM-v7.0-PKU), developed at Peking University. Could you please review this model registration?

Users

Name: Fan Wu
Email: fwu25@stu.pku.edu.cn
GitHub: FanWu1020

Name: Yang Ou
Email: yang.ou@pku.edu.cn
GitHub: ouyang363


danielhuppmann:
Thank you for submitting the model registration via PR.

I assume that this region-definition and mapping is identical to GCAM 7.0, see [here](https://github.com/IAMconsortium/common-definitions/blob/main/definitions/region/native_regions/GCAM/GCAM_7.0.yaml)?

In this case, it is not necessary to add a complete definition and mapping.

For consistency with other GCAM model versions, I suggest to use "GCAM-PKU 7.0" as model name (see [here](https://github.com/IAMconsortium/common-definitions/blob/main/mappings/GCAM/GCAM_8.2.yaml) for a similar case). Please confirm that this is ok for you.

FanWu1020:
Hi Daniel @danielhuppmann,

Thank you for the clarification. We have updated the PR accordingly:
GCAM-PKU 7.0 now reuses the existing GCAM 7.0 region definition and mapping.

Fan

PR #430, opened 2026-09-01
Register SuCCESs IAM, version 2026-08-31

tekholm:
Add SuCCESs model mapping: mappings/SuCCESs_2026-08-31.yaml

SuCCESs is a global, single region IAM that models energy, land-use, materials and climate, developed at FMI.
GitHub: https://github.com/SuCCESsIAM/SuCCESsIAM
Model description: https://gmd.copernicus.org/articles/18/4805/2025/

For review: @danielhuppmann @phackstock


phackstock (approved):
Thanks for adding @tekholm, looks good to me.

PR #431, opened 2026-09-02
Registering the REMIND-MAgPIE 3.4-4.9 model

gabriel-abrahao:
New REMIND-MAgPIE version for @merfort 's Land Matters paper. Could you please take a look @danielhuppmann @phackstock ? Thanks!

### DIFF
--- a/mappings/REMIND-MAgPIE/REMIND-MAgPIE_3.4-4.9.yaml
+++ b/mappings/REMIND-MAgPIE/REMIND-MAgPIE_3.4-4.9.yaml
@@ -1,0 +1,111 @@
+model:
+  - REMIND-MAgPIE 3.4-4.9
+  - REMIND-MAgPIE 3.4.0-4.9.0
+
+native_regions:
+  - CAZ: REMIND-MAgPIE 3.4-4.9|Canada, Australia, New Zealand
+  - CHA: REMIND-MAgPIE 3.4-4.9|China and Taiwan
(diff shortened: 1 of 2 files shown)


phackstock:
Thanks for adding this @gabriel-abrahao. Quick question from my side since the version number of the model seems to be an earlier one compared to the latest REMIND-MAgPIE versions, just to make sure that this is intentional.

gabriel-abrahao:
Thanks for checking, Philip, but that's intentional indeed. This is for a paper that is about to be published and, unsurprisingly, that means the runs were made quite a while ago 🙃 . Also, I don't know if this is with you, but I seem to not have the proper authorization to submit scenarios to the SCI sandbox. Could you give me access? The email is abrahao@pik-potsdam.de

Thanks again!

mappings/REMIND-MAgPIE/REMIND-MAgPIE_3.4-4.9.yaml
```diff
@@ -0,0 +1,111 @@
+model:
+  - REMIND-MAgPIE 3.4-4.9
+  - REMIND-MAgPIE 3.4.0-4.9.0
```
danielhuppmann, line 3: Not sure if you really want the `.0` version here as well? This seems to be only ever used for version `REMIND-MAgPIE 3.2.0-4.8.0`... Consider removing this in a follow-up PR here and in the other mapping file.

danielhuppmann (approved):
Looks good, thank you!

PR #437, opened 2026-09-12
Register EPPA 8

danielhuppmann:
FYI @chenyhmitedu

I transferred the definition and mapping file from https://github.com/chenyhmitedu/EPPA and made the following changes:
- Harmonise country names with nomenclature country names (see https://docs.ece.iiasa.ac.at/standards/countries.html)
- Remove aggregates like "South Central Africa" - please provide country explicit names (or we merge as is)
- Harmonise region-display region names between definitions and mapping
- Apply formatting (remove tabtops and enforce line breaks)

You can look at the individual commits for the detailed changes.

### DIFF
--- a/mappings/EPPA_v8.yaml
+++ b/mappings/EPPA_v8.yaml
@@ -1,0 +1,146 @@
+model:
+  - EPPA 8
+
+native_regions:
+  - USA: EPPA 8|United States
+  - CAN: EPPA 8|Canada
+  - MEX: EPPA 8|Mexico
+  - JPN: EPPA 8|Japan
+  - ANZ: EPPA 8|Australia - New Zealand
+  - EUR: EPPA 8|Europe
+  - ROE: EPPA 8|Rest of Europe
+  - RUS: EPPA 8|Russia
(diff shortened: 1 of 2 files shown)


mappings/EPPA_v8.yaml
```diff
@@ -0,0 +1,146 @@
...
+  - India:
+      - IND
+  - Japan:
+      - JPN
+  - South Korea:
+      - KOR
+  - United States:
+      - USA
+  - African Union:
+      - AFR
```
danielhuppmann, line 146: We could add EU & Uk explicitly for easier comparison between models (Switzerland and Norway are included, but that is probably close enough), @chenyhmitedu 
```suggestion
      - AFR
  - European Union and United Kingdom:
      - EUR
```

chenyhmitedu:
Thanks a lot Daniel! All your suggestions and revisions are well taken. Per "Soutch Central Africa", please replace it by the following two countries:
Angola
São Tomé and Príncipe

danielhuppmann:
Thank you, added the EU & UK aggregate and the two countries.

PR #438, opened 2026-09-12
Register GEMINI v7.3

danielhuppmann:
Adding the model registration files for GEMINI 7.3 received via email - please take a look @mvielle 

The second commit implemented hHarmonise country names with nomenclature country names (see https://docs.ece.iiasa.ac.at/standards/countries.html)

Also, using aggregates in the `countries` (e.g., "Rest of East Asia" or "Caribbean") is not allo

It would also be helpful to replace the abbreviations in the native-region-names (as used in the Scenario Explorer) like `GEMINI 7.3|ASI` with more intuitive and readable names, e.g. `GEMINI 7.3|Asia`. Any objections?

FYI @MathijsHarmsenPBL

### DIFF
--- a/definitions/region/native_regions/GEMINI_v7.3.yaml
+++ b/definitions/region/native_regions/GEMINI_v7.3.yaml
@@ -1,0 +1,169 @@
+- GEMINI 7.3:
+  - GEMINI 7.3|ASI:
+      countries:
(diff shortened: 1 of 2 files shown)


danielhuppmann:
@mvielle to review by end of September

PR #440, opened 2026-09-16
Register AIM-Technology 2.0

danielhuppmann:
FYI @kenoshiro 

This registration was parsed from a model registration xlsx file. The mapping for G20 countries/regions was added manually (see the second commit).


danielhuppmann:
Waiting for confirmation from Tomoko Hasegawa if this should be used also for `AIM 2.0`, etc.

tmkhsgw:
No. It should not be used for AIM 2.0 etc. AIM-Technology 2.0 and other AIM models e.g. AIM-Hub 2.0 have different regions.

danielhuppmann:
Thank you for confirming!

